PR #8, opened 2025-12-27
fix(quota): filter out unknown percentages when calculating lowest quota

nguyenphutrong:
## Summary
- Fix incorrect quota notifications triggered when percentage is -1 (unknown)
- Filter out invalid percentages (-1) when calculating lowest quota for menu bar display
- Affects Cursor and Gemini CLI providers which use -1 for unknown/unavailable quota

## Problem
When providers like Cursor or Gemini CLI don't have quota data, `percentage` is set to `-1` (unknown). The old logic used `.min()` to get the lowest value, causing `-1` to be selected and triggering false quota alerts (since `-1 <= 20` is true).

## Changes
- `QuotaViewModel.swift` - `checkQuotaNotifications()`: Filter `{ $0 >= 0 }` to exclude unknown percentages
- `QuotaViewModel.swift` - `menuBarQuotaItems`: Same filter for menu bar display
- `QuotioApp.swift` - `quotaItems`: Same filter for menu bar display

## Testing
- Build succeeded ✅

### DIFF
--- a/Quotio/QuotioApp.swift
+++ b/Quotio/QuotioApp.swift
@@ -42,7 +42,9 @@ struct QuotioApp: App {
             if let accountQuotas = viewModel.providerQuotas[provider],
                let quotaData = accountQuotas[selectedItem.accountKey],
                !quotaData.models.isEmpty {
-                let lowestPercent = quotaData.models.map(\.percentage).min() ?? 0
+                // Filter out -1 (unknown) percentages when calculating lowest
+                let validPercentages = quotaData.models.map(\.percentage).filter { $0 >= 0 }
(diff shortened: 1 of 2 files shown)


ganuong11:
Confirmed merged and shipped in **v0.2.2**. The `{ $0 >= 0 }` filter excluding unknown (`-1`) percentages is applied in all three lowest-quota paths (`QuotaViewModel.checkQuotaNotifications`, `QuotaViewModel.menuBarQuotaItems`, `QuotioApp.quotaItems`), which stops the false low-quota alerts for Cursor / Gemini CLI.

Note: the same unknown-percentage guard was later extended to Antigravity `GroupedModelQuota` in a follow-up commit (`e1561e7`, also referencing this PR). Closing as complete.

PR #119, opened 2026-01-06
fix(agent-config): adjust footer height in AgentConfigSheet to improve layout

pakholeung37:
Adjusts the footer height in AgentConfigSheet from 800 to 640 to improve layout on smaller screens.


pakholeung37:
image

Issue #148, opened 2026-01-11
Proxy fails to start due to duplicate glm-api-key entries in config file


totally-tim:
### What happened?

Issue: Proxy fails to start due to duplicate glm-api-key entries in config file
Description
The proxy fails to start silently when a GLM custom provider is configured. The root cause is that glm-api-key: entries accumulate in the config file instead of being replaced on each sync, resulting in invalid YAML with duplicate keys.

FIX: Add "glm-api-key:" to both hardcoded lists in CustomProviderService.swift lines 195 and 220, OR better yet, derive the list dynamically from CustomProviderType.allCases.map { "\($0.rawValue):" }.

### Steps to reproduce

1. Add a custom provider of type "GLM Compatible"
2. Start the proxy
3. Stop and start the proxy again (or trigger any action that calls syncCustomProvidersToConfig())
4. Repeat step 3 several times
5. Observe that the proxy no longer starts and the config.yaml has a ton of api secrets:
<img width="250" height="766" alt="Image" src="https://github.com/user-attachments/assets/350509b6-f877-4107-8f94-2820b377b839" />

### Expected behavior

Proxy starts normal when clicking start AND no excess api secrets in config.yaml

### Severity

🟠 High - Core feature broken, no workaround

### Affected component

Proxy Server (CLIProxyAPI)

### Provider (if applicable)

Custom Provider

### CLI Agent (if applicable)

N/A

### macOS Version

26.2 (25C56) Tahoe

### Quotio Version

0.7.0 (32)

### App Mode

Full Mode (Proxy running)

### Logs (optional)

```shell
[RequestTracker] Loaded 2 entries from disk
[CloudflaredService] Cleaned up orphan cloudflared processes
[CloudflaredService] Cleaned up orphan cloudflared processes
[CLIProxyManager] Cleaned up orphan processes on ports 8317 and 18317
nw_connection_copy_protocol_metadata_internal_block_invoke [C8] Client called nw_connection_copy_protocol_metadata_internal on unconnected nw_connection
nw_connection_copy_protocol_metadata_internal_block_invoke [C8] Client called nw_connection_copy_protocol_metadata_internal on unconnected nw_connection
nw_connection_copy_connected_local_endpoint_block_invoke [C8] Client called nw_connection_copy_connected_local_endpoint on unconnected nw_connection
nw_connection_copy_connected_remote_endpoint_block_invoke [C8] Client called nw_connection_copy_connected_remote_endpoint on unconnected nw_connection
[...]

PR #149, opened 2026-01-11
Fix duplicate glm-api-key entries in config file (#148)

totally-tim:
## Summary
based on my issue: https://github.com/nguyenphutrong/quotio/issues/148 - Proxy fails to start due to duplicate `glm-api-key` entries accumulating in the config file.
## Problem
The `removeCustomProviderSections()` function used a hardcoded list of custom provider keys that did not include `glm-api-key:`. This caused GLM provider configurations to accumulate rather than being replaced on each sync, resulting in invalid YAML that CLIProxyAPI could not parse.
## Solution
Replace the hardcoded list with a dynamically derived list from `CustomProviderType.allCases`. This ensures all provider types (including `glm-api-key` and any future types) are properly cleaned up before syncing.
## Changes
- `CustomProviderService.swift`: Use `CustomProviderType.allCases.map { "\($0.rawValue):" }` instead of hardcoded array
## Testing
- Built and verified the fix compiles successfully
- Users with corrupted configs should manually clean up duplicate entries or delete and recreate their config

### DIFF
--- a/Quotio/Services/CustomProviderService.swift
+++ b/Quotio/Services/CustomProviderService.swift
@@ -186,13 +186,15 @@ final class CustomProviderService {
     private func removeCustomProviderSections(from content: String) -> String {
         var result = content
         
+        // Dynamically derive custom provider keys from CustomProviderType enum
+        // This ensures new provider types are automatically handled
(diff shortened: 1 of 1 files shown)


nguyenphutrong:
Thank you. I’ll merge and release it soon.

Issue #151, opened 2026-01-11
Ignoring "extra usage" of Anthropic Account in usage calculation


totally-tim:
### Problem Statement

Currently if your extra usage within anthropic is zero or has been reached, the full usage bar shows no usage remaining, even if the actual session is available (screenshot shows it well)

<img width="1282" height="304" alt="Image" src="https://github.com/user-attachments/assets/18c8a98d-b4a5-47d9-bb4e-e3dfd3bfb4e7" />

### Proposed Solution

Usage should reflect actual remaining session usage, not extra usage

### Alternative Solutions

potentially a mix? Maybe "if session usage 0%, count extra usage as calculation"

### Feature Category

Quota Tracking Enhancement

### How important is this to you?

🟡 Medium - Nice to have

### Use Case

As a User, I would like my modals to correctly reflect my available usage for an claude anthropic account

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I am willing to contribute to implementing this feature

totally-tim:
to add onto this, i how should model specific session limits be defined regarding "total usage remaining"?
example:

<img width="1288" height="367" alt="Image" src="https://github.com/user-attachments/assets/f8b91007-438a-4310-a2b0-d46ab57130bf" />
It seems like currently it just counts the "worst" out of session limits.

Maybe a toggle/setting in the setting should allow a user to define if "total usage" is defined by included usage + extra usage or only included usage? (and model specific usage also maybe being an average of all models or the least available model like now)

PR #160, opened 2026-01-12
Add customizable usage calculation and aggregation modes

totally-tim:
Based on issue: https://github.com/nguyenphutrong/quotio/issues/151
Introduces settings for total usage calculation (session-only vs combined) and model aggregation (lowest vs average) in MenuBarSettings. Updates all quota displays to use these new modes, adds a Usage Display section in Settings, and refactors related logic for consistency across the app.
Settings:
image

image

image

and (like how it used to be) lowest remaining:
image

image


totally-tim:
Feature is pretty feature complete, I just want to do some more manual testing and would appreciate of others also test it out!

PR #166, opened 2026-01-12
Fix fallback model picker to list all providers

totally-tim:
Based on explanation in https://github.com/nguyenphutrong/quotio/discussions/150

Currently the model selection feature is not correctly fetching all models for each provider:
image

My fix is bringing back all authenticated providers and their models to be selected:
image

image

## Summary
- load fallback models per auth file via management API
- avoid collapsing identical model IDs across providers
- refresh fallback model list when proxy starts or entry sheet opens
## Testing
- Manual: opened Fallback → Add Entry and confirmed models list across providers

### DIFF
--- a/Quotio/Views/Components/FallbackSheets.swift
+++ b/Quotio/Views/Components/FallbackSheets.swift
@@ -1,37 +1,30 @@
 //
 //  FallbackSheets.swift
-//  Quotio - Fallback Configuration Sheets
+//  Quotio - Model Fallback Configuration
 //
 
 import SwiftUI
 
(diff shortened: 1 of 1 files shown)


totally-tim:
current bug: copilot provider when selecting Opus 4.5 is showing up as anthropic in fallback list:
(in my example I selected antigravity -> anthropic -> copilot:
image

totally-tim:
> current bug: copilot provider when selecting Opus 4.5 is showing up as anthropic in fallback list: (in my example I selected antigravity -> anthropic -> copilot: image

fixed:

image

PR #288, opened 2026-02-07
fix: prevent Quotio from killing itself during port cleanup

benzntech:
## Summary
- Fixes crash when clicking "Apply Workaround (Backup & Force URL)" in Settings > Troubleshooting
- The app would silently disappear because `killProcessOnPortSync()` was killing Quotio's own process
- When ProxyBridge is enabled (default), Quotio owns the proxy port via in-process `NWListener`
- Added check to skip own PID to prevent self-termination

## Root Cause
When restarting the proxy, `killProcessOnPortSync()` uses `lsof` to find PIDs on the port, then kills them. Due to a race condition where ProxyBridge stops but port release is asynchronous, `lsof` would return Quotio's own PID. The subsequent `kill(pid, SIGKILL)` terminated Quotio itself (SIGKILL doesn't generate crash reports, hence "silent disappearance").

## Test plan
- [ ] Launch Quotio
- [ ] Go to Settings > Troubleshooting
- [ ] Click "Apply Workaround (Backup & Force URL)"
- [ ] Verify app remains running and proxy restarts successfully

### DIFF
--- a/Quotio/Services/Proxy/CLIProxyManager.swift
+++ b/Quotio/Services/Proxy/CLIProxyManager.swift
@@ -1061,25 +1061,34 @@ final class CLIProxyManager {
     
     /// Synchronous port cleanup for use in detached tasks.
     /// This method is `nonisolated` to allow calling from background threads.
+    /// IMPORTANT: Excludes own PID to prevent killing Quotio itself when ProxyBridge is running.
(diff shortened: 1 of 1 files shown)


benzntech:
Closing as superseded by #290 which includes this fix plus additional improvements.

PR #307, opened 2026-02-11
fix(proxy): persist upstream proxy URL across restarts

hongxunming:
## Summary

- **Persist proxy URL to UserDefaults** when saved from the Settings UI, so the value survives app/proxy restarts
- **Guard `syncProxyURLInConfig()`** to skip overwriting `config.yaml` when no local preference has been set, preserving values configured via the management UI or direct file editing

## Problem

The upstream proxy URL was lost on every restart because:
1. `saveProxyURL()` only called the proxy API but never persisted to `UserDefaults`
2. On restart, `syncProxyURLInConfig()` read an empty `UserDefaults` value and overwrote `config.yaml` with nil

This affected all three configuration methods: Quotio UI, management web UI (`/management.html`), and direct `config.yaml` editing.

## Steps to Reproduce

1. Quotio Settings → Upstream Proxy → paste `socks5://127.0.0.1:7890` → press Enter → switch to another screen and back → URL is cleared
2. Set proxy URL via management UI (`http://127.0.0.1:18317/management.html#/config`) → restart Quotio → URL is cleared
3. Edit `~/Library/Application Support/Quotio/config.yaml` `proxy-url` → restart → value disappears

## Changes

| File | Change |
|------|--------|
| `SettingsScreen.swift` | Save proxy URL to `UserDefaults` before calling the API |
| `CLIProxyManager.swift` | Early return in `syncProxyURLInConfig()` when `UserDefaults` key has never been set |

## Test plan

- [ ] Set proxy URL in Quotio Settings → press Enter → restart app → verify URL persists
- [ ] Set proxy URL via management UI → restart app → verify URL persists
- [ ] Edit `config.yaml` `proxy-url` directly → restart app → verify URL persists
- [ ] Clear proxy URL in Quotio Settings → restart → verify it stays empty
- [ ] Fresh install (no `proxyURL` in UserDefaults) → set URL via management UI → verify no overwrite on restart

### DIFF
--- a/Quotio/Services/Proxy/CLIProxyManager.swift
+++ b/Quotio/Services/Proxy/CLIProxyManager.swift
@@ -528,6 +528,12 @@ final class CLIProxyManager {
         }
     }
     
+    /// Synchronizes the proxy URL from UserDefaults into the config.yaml file.
+    ///
+    /// If no local preference has been set (i.e., the user has never saved a proxy URL from the UI),
+    /// this method returns early to preserve any value configured via the management web UI or
+    /// direct config.yaml editing. Once the user sets a value through the app, UserDefaults becomes
+    /// the source of truth and its value is written to config.yaml on each proxy start.
     private func syncProxyURLInConfig() {
         // If no local preference exists, keep config.yaml as-is (it may be set via management UI).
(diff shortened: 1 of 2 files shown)


hongxunming:
#286

Issue #302, opened 2026-02-10
Antigravity opus 4.6 Display Error in Fallback Model Selection


hongxunming:
### What happened?

In the Fallback Model configuration, selecting the Antigravity model incorrectly results in it being displayed as a Claude model.

### Steps to reproduce

1. **Fallback Model Interface:** Navigate to "Add," select the **Anthropic Claude Opus 4.6 Thinking** model, and click "Confirm." The interface then displays the model type as "Claude."

https://github.com/user-attachments/assets/e9c75266-46a5-474a-9ba0-6e30d7e991fd

2. **Log Interface:** The logs also incorrectly identify and display the **Claude Opus 4.6 Thinking** model as a "Claude" type.

<img width="2458" height="114" alt="Image" src="https://github.com/user-attachments/assets/61f7da51-a1f7-4f86-92d6-feb03072638f" />

### Expected behavior

The system should correctly identify and display the model as the Antigravity type in both the configuration and log interfaces.

### Severity

🟢 Low - Minor issue, cosmetic

### Affected component

Other

### Provider (if applicable)

Antigravity

### CLI Agent (if applicable)

None

### macOS Version

26.2

### Quotio Version

0.10.0

### App Mode

Full Mode (Proxy running)

### Logs (optional)

```shell

```

Junrin-Lee:
GitHub Copilot has the same issue. I've already fixed it in #293. 

Could you please let me know if I can submit a PR to address this? If so, please assign this issue to me. 

I'm not entirely sure if this is a bug or something else I'm not aware of.

@nguyenphutrong

hongxunming:
Issue and Resolution Summary

The issue was caused by the CLIProxy passing the model name claude-opus-4-6-thinking, which led Quotio to incorrectly identify the provider as "Claude." 

To resolve this, a model alias must be configured to ensure the correct routing. 

Specifically, you need to navigate to the management console at http://127.0.0.1:18317/management.html#/auth-files/oauth-model-alias?provider=antigravity and add a mapping to alias claude-opus-4-6-thinking to gemini-claude-opus-4-6-thinking.

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/c65d6722-dcdb-430a-8acd-12179ec2de61" />

Junrin-Lee:
Saddly, I tried it but not work for github-copilot 😭 but work well in antigravity 🤔

[...]

PR #308, opened 2026-02-11
fix(fallback): fix provider icons, context menu scope, and add model refresh

hongxunming:
## Summary

Two bug fixes and one feature enhancement in the Fallback UI:

### 1. Aggregator provider icons misidentified

- Aggregator providers (Copilot, Antigravity, Kiro) host models from other providers,
  so model IDs carry the underlying provider's prefix (e.g., Copilot's `claude-3.5-sonnet`)
- `providerFromModel()` checked model ID prefix first, causing aggregator-hosted models
  to display the wrong provider icon
- **Fix**: Check the `provider` field (from proxy API `owned_by`) first, fall back to
  model ID prefix matching for unknown provider metadata

| Model ID | Provider field | Before (❌) | After (✅) |
|---|---|---|---|
| `claude-3.5-sonnet` | `github-copilot` | Claude icon | Copilot icon |
| `claude-opus-4-6-thinking` | `antigravity` | Claude icon | Antigravity icon |

### 2. Context menu deleting entire virtual model group

- `.contextMenu` was attached to the entire `DisclosureGroup`, covering both the title row
  and expanded fallback entry rows
- Right-clicking a fallback entry triggered group-level context menu instead of entry-level one
- **Fix**: Move `.contextMenu` to the label's `HStack` so only the title row responds to right-click

### 3. Add refresh button to AddFallbackEntry sheet

**Background**: When users open the "Add Fallback Entry" sheet, the model list is loaded once from the proxy on initial screen load. If the proxy starts after the screen is already open, or if new models become available, users have to close and reopen the entire Fallback settings to get an updated list — with no indication that a refresh is even possible.

**What this does**: Adds a compact refresh button next to the model picker (and the empty-state hint) so users can reload the model list without leaving the sheet.

- Button is only visible when the proxy is running (no point refreshing with no backend)
- Shows a smooth spinning animation during the fetch
- Displays a brief ✅ / ❌ icon on success/failure before returning to idle
- `AgentSetupViewModel.loadModels()` now returns `Bool` indicating fetch success (`@discardableResult` keeps existing call sites unchanged)

https://github.com/user-attachments/assets/97f6d323-7426-4496-889a-357089eb8a33

## Changed files

- `Quotio/Views/Components/FallbackSheets.swift` — reorder resolution steps in `providerFromModel()`, add refresh button with `TimelineView`-driven rotation and success/failure feedback
- `Quotio/Views/Screens/FallbackScreen.swift` — narrow context menu scope in `VirtualModelRow`, conditionally provide `onRefresh` closure based on proxy status
- `Quotio/ViewModels/AgentSetupViewModel.swift` — `loadModels()` returns `Bool`
- `Quotio/Localizable.xcstrings` — localization updates

## Test plan

**Provider icon fix**
- [x] Aggregator-hosted models display the aggregator's icon correctly
- [x] Direct provider models still display the correct icon

**Context menu fix**
- [x] Right-clicking a fallback entry row does not show the virtual model context menu
- [x] Right-clicking virtual model title row still shows context menu normally

**Refresh button**
- [x] Refresh button appears next to model picker when proxy is running
- [x] Refresh button is hidden when proxy is not running
- [x] Clicking refresh shows spinning animation, then ✅ on success
- [ ] Stopping proxy mid-refresh shows ❌ feedback
- [x] Model list updates in-place after successful refresh
- [x] Previously selected model is preserved if still valid after refresh

**General**
- [x] Local build succeeded and app runs without issues

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -8609,6 +8609,35 @@
         }
       }
     },
+    "menubar.hideAccounts" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Hide accounts"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Masquer les comptes"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Ẩn tài khoản"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "隐藏账户"
+          }
+        }
+      }
+    },
     "menubar.hideFromMenuBar" : {
       "extractionState" : "stale",
       "localizations" : {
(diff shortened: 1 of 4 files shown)


hongxunming:
#293
#302

PR #322, opened 2026-02-14
feat: add model selection and connection testing for custom providers

benzntech:
## Summary
- Reorder custom provider form: Base URL/API Key first, then model selection (Step 1-3)
- Add "Fetch Models" button to fetch available models from `/v1/models` endpoint
- Show top 5 popular models first, with search box to filter all models
- Add Select All / Deselect All buttons for easy model selection
- Add "Limit models to selection" toggle - when enabled, only selected models are exposed via proxy
- Test API connection before saving provider configuration

## Problem
When adding a custom provider like OpenRouter (200+ models), users had no way to limit which models are available via the proxy. All models were exposed, making it confusing to select specific models.

## Solution
1. **Reordered form flow**: Provider info → API Key → Model selection (logical progression)
2. **Model fetching**: Fetch models from provider's `/v1/models` endpoint
3. **Visual selection**: Checkboxes with search, top 5 popular models highlighted
4. **Limit toggle**: Option to expose only selected models via proxy
5. **Connection testing**: Validates API key works before saving

## Testing
- Added custom provider (OpenRouter, Ollama) with API key
- Fetched models successfully from provider endpoint
- Selected specific models and verified they're saved to config
- Tested invalid API key - shows error message

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -4462,6 +4462,136 @@
         }
       }
     },
+    "customProviders.fetchModels" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Fetch Models"
+          }
+        }
+      }
+    },
+    "customProviders.allModels" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "All Models"
+          }
+        }
+      }
+    },
+    "customProviders.popularModels" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Popular Models"
+          }
+        }
+      }
+    },
+    "customProviders.searchModels" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Search models..."
+          }
+        }
+      }
+    },
+    "customProviders.selectedModels" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%lld models selected"
+          }
+        }
+      }
+    },
+    "customProviders.clearSelection" : {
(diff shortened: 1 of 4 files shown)


nguyenphutrong:
## Review: Needs fixes before merge

### 🚨 CRITICAL: Breaking Change - Backward Compatibility

The new `limitToSelectedModels: Bool` property added to `CustomProvider` struct (line 234 in `CustomProviderModels.swift`) breaks JSON decoding for existing stored providers.

**Problem**: All existing users will lose their custom provider data on update because:
1. `UserDefaults` contains JSON of custom providers without the `limit-to-selected-models` key
2. Auto-synthesized `Codable` throws when key is missing
3. `catch` block in `CustomProviderService.loadProviders()` resets `providers = []`

**Fix required**:
```swift
// Add custom init(from:) to CustomProvider
init(from decoder: Decoder) throws {
    let container = try decoder.container(keyedBy: CodingKeys.self)
    // ... existing properties ...
    self.limitToSelectedModels = try container.decodeIfPresent(
        Bool.self, forKey: .limitToSelectedModels
    ) ?? true  // Default to true for existing data
}
```

### 🔴 HIGH: Connection test is mandatory on save

`saveProvider()` always runs connection test before saving. Users cannot save config when:
- Network is down
- API temporarily unavailable
- Just want to save and test later

**Fix**: Add "Skip Test" button or make test optional.

### 🟡 Medium issues:
1. **Code duplication**: `fetchModelsFromAPI()` and `testConnection()` share ~40 identical lines - should extract common logic
2. **URL path bug**: `url.appendingPathComponent("v1/models")` can create `/v1/v1/models` if base URL already ends with `/v1`
3. **Hardcoded models**: Popular model IDs (`gpt-4o`, `claude-3-opus`) will become outdated
4. **Claude compatibility**: Button appears for Claude type but Claude API doesn't have `/v1/models` endpoint

Please address the CRITICAL issue before merge.

f4cker:
The `HIGH: Connection test is mandatory on save` seems still not be fixed.

PR #328, opened 2026-02-15
fix: prevent menu bar from dropping accounts

jakozloski:
## Summary
- Raise default max menu bar items from 3 to 5 and auto-expand when new accounts are discovered (up to hard cap of 10)
- Remove `hasUserModifiedMenuBar` guard that blocked auto-selection of new accounts after any manual toggle
- Preserve existing Codex quota data when the OpenAI API fetch fails (matching Claude fetcher pattern)
- Skip CLI binary existence check in dropdown when quota data already exists for the provider

## Test plan
- [ ] Add a new Claude/Codex auth file and verify it auto-appears in menu bar and dropdown
- [ ] Verify existing accounts persist across app restarts even when API calls intermittently fail
- [ ] Confirm Codex account shows in dropdown when `codex` binary is in a non-standard path
- [ ] Test with >5 accounts to verify auto-expand works up to the 10-item cap

### DIFF
--- a/Quotio/Models/MenuBarSettings.swift
+++ b/Quotio/Models/MenuBarSettings.swift
@@ -440,7 +440,7 @@ final class MenuBarSettingsManager {
 
     static let minMenuBarItems = 1
     static let maxMenuBarItems = 10
-    static let defaultMenuBarMaxItems = 3
+    static let defaultMenuBarMaxItems = 5
 
     /// Whether to show menu bar icon at all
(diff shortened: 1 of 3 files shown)


jakozloski:
Closing - these changes are already covered by our fork's PR #1 (commits 52f4027 and 6a0bef4 which removed the menu bar cap, flattened the account list, and fixed the OpenAI percentage display).

PR #325, opened 2026-02-14
feat: add copy URL option for OAuth login flow

benzntech:
## Summary
- Add auth URL storage in OAuthState for copy functionality
- Show "Copy Link" and "Open Link" buttons during OAuth polling
- User can now copy auth URL before opening in browser
- Matches EasyCLI functionality for OAuth flow

## Changes
- `QuotaViewModel.swift`: Added `authURL` property to `OAuthState`, store URL instead of auto-opening
- `ProvidersScreen.swift`: Updated `OAuthStatusView` to show copy/open buttons when URL is available
- `Localizable.xcstrings`: Added localization strings for `oauth.copyLinkOrOpen`, `oauth.copyLink`, `oauth.openLink`

## Usage
1. Click "Authenticate" for a provider
2. During polling phase, see "Copy Link" and "Open Link" buttons
3. Copy URL to clipboard or open directly in browser

## Tested
- Build succeeded

PR #330, opened 2026-02-16
fix: don't auto-open browser during OAuth flow

benzntech:
## Summary
- Remove automatic browser opening during OAuth flow
- User must click "Open Link" button to open auth URL in browser

## Issue
Fixes issue from #325 where the browser was opening automatically before user could click the "Open Link" button. The original PR added copy/open buttons but still auto-opened the browser, defeating the purpose of having user control.

## Changes
- `QuotaViewModel.swift`: Removed `NSWorkspace.shared.open(url)` call in `startOAuth()`

## Testing
- Build succeeded

### DIFF
--- a/Quotio/ViewModels/QuotaViewModel.swift
+++ b/Quotio/ViewModels/QuotaViewModel.swift
@@ -1378,11 +1378,7 @@ final class QuotaViewModel {
                 return
             }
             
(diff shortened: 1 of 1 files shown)


owenmead:
This breaks the oauth flow when having to refresh the auth. Tested with Claude Code. No url is given to complete auth.

PR #353, opened 2026-03-08
Add GPT-5.4 and remove fake GPT-5.4 Codex variants

Vic563:
## Summary

This keeps plain `gpt-5.4` available in Quotio while removing the fake `gpt-5.4-codex` and `gpt-5.4-codex-mini` variants.

## Changes

- keeps built-in `gpt-5.4`
- removes built-in `gpt-5.4-codex` and `gpt-5.4-codex-mini`
- normalizes stale saved selections:
  - `gpt-5.4-codex` -> `gpt-5-codex`
  - `gpt-5.4-codex-mini` -> `gpt-5-codex-mini`
- sanitizes fetched and fallback model lists so removed aliases do not reappear in the UI

## Why

`gpt-5.4` is a real model entry we want to expose, but the `gpt-5.4-codex*` variants are not real and should not appear as selectable models.

## Testing

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build`

### DIFF
--- a/Quotio/Models/AgentModels.swift
+++ b/Quotio/Models/AgentModels.swift
@@ -222,6 +222,12 @@ nonisolated struct AvailableModel: Identifiable, Codable, Hashable, Sendable {
     let provider: String
     let isDefault: Bool
 
+    static let legacyModelReplacements: [String: String] = [
+        "gpt-5.4": "gpt-5",
+        "gpt-5.4-codex": "gpt-5-codex",
+        "gpt-5.4-codex-mini": "gpt-5-codex-mini"
+    ]
+
     var displayName: String {
         name.split(separator: "-")
             .map { $0.capitalized }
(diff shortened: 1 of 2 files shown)


nguyenphutrong (changes requested):
Thanks for the focused cleanup. The branch merges cleanly into the current `master`, and both the Debug build and test suite pass. There is still a blocking model-identity issue before this is safe to merge:

- `VirtualModel.name` is an arbitrary, trim-only user value and fallback resolution later compares it by exact string. `refreshVirtualModels()` therefore exposes a virtual model named `gpt-5.4-codex`, but `updateModelSlot()` stores it as `gpt-5-codex`; the restore paths perform the same rewrite. The request no longer matches the virtual model, so its fallback chain is silently bypassed. Please scope legacy migration to known upstream model IDs, or explicitly reserve/migrate virtual-model names, rather than normalizing every raw model string.
- This domain distinction also needs to resolve the fallback-list requirement: appending virtual names verbatim allows those aliases to remain visible, while normalizing them would corrupt their identity. Please define whether these names are reserved and enforce that consistently at creation/rename/restore, or preserve custom virtual names while sanitizing only upstream models.
- `sanitizedModels(_:)` keeps whichever normalized ID appears first. When both a legacy alias and its canonical ID are returned, the result is order-dependent and can discard canonical metadata. Prefer the canonical entry deterministically and avoid deduplicating unrelated IDs as part of this migration.

Please add focused regression coverage for saved upstream selections, arbitrary virtual-model names, and canonical/legacy ordering. The existing suite does not exercise these paths.

nguyenphutrong:
Thank you for taking the time to prepare this contribution. After reviewing it against the current codebase, we've decided not to move forward with this PR. The global model-name normalization conflicts with user-defined virtual model identities, the sanitization behavior remains inconsistent for fallback models, and the deduplication can produce order-dependent results. The supported model landscape has also evolved significantly since this PR was opened, so addressing the current requirements in a fresh, narrowly scoped change would be safer than continuing this implementation. Thank you again for the effort and for helping improve Quotio.

PR #415, opened 2026-05-30
feat: CLIProxyApi update

Luzenrocy:
1. Add CLIProxyApi update functionality to the Settings page.
2. Add CLIProxyApi update logic to the Check for Updates button in the menu bar.

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -17459,6 +17459,24 @@
         }
       }
     },
+    "settings.proxyUpdate.alreadyLatest" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : { "stringUnit" : { "state" : "translated", "value" : "CLIProxyAPI is already up to date" } },
+        "fr" : { "stringUnit" : { "state" : "translated", "value" : "CLIProxyAPI is already up to date" } },
+        "vi" : { "stringUnit" : { "state" : "translated", "value" : "CLIProxyAPI is already up to date" } },
+        "zh-Hans" : { "stringUnit" : { "state" : "translated", "value" : "CLIProxyAPI 已是最新版本" } }
+      }
+    },
+    "settings.proxyUpdate.checkAndUpdate" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : { "stringUnit" : { "state" : "translated", "value" : "Check and Update CLIProxyAPI" } },
+        "fr" : { "stringUnit" : { "state" : "translated", "value" : "Check and Update CLIProxyAPI" } },
+        "vi" : { "stringUnit" : { "state" : "translated", "value" : "Check and Update CLIProxyAPI" } },
+        "zh-Hans" : { "stringUnit" : { "state" : "translated", "value" : "检查并更新 CLIProxyAPI" } }
+      }
+    },
     "settings.proxyUpdate.checkNow" : {
       "extractionState" : "stale",
       "localizations" : {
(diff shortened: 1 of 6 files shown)


nguyenphutrong (changes requested):
Requesting changes for the following correctness issues:

1. [P1] Pin and revalidate the binary source across suspension points (CLIProxyManager.swift:1663-1685). The method checks selectedBinarySource before awaiting the GitHub request, then reads it again when constructing ProxyVersionInfo and resolving the local version. Since MainActor tasks are reentrant, a user can switch to plusLocal during the request. The upstream asset is then labeled as plusLocal, the source guard in performManagedUpgrade passes, and the upstream binary can be installed and promoted into the bundled track. Snapshot the source before the first await, use that snapshot throughout, and abort if the selection changes before installation.

2. [P1] Serialize proxy update operations at the manager or service level (UpdaterService.swift:118-135 and CLIProxyManager.swift:1683-1685). Every invocation of the now-always-enabled menu command creates a new untracked Task. performManagedUpgrade leaves managerState as idle or active while downloading, so two invocations can both pass its guard and later overwrite shared testProcess, testPort, and testManagementKey state. A view-local isUpgrading flag is not sufficient because the menu and other update surfaces remain independent. Use a shared in-flight task or set an operation state synchronously before the first await.

3. [P2] Do not install a missing local proxy from the generic app update command (UpdaterService.swift:118-123 and CLIProxyManager.swift:1675-1685). When both currentVersion and installedProxyVersion are nil, shouldUpdate returns true by default. Therefore Check for Updates can silently download and promote CLIProxyAPI even in Remote Proxy Mode or when no local proxy was installed. Restrict automatic proxy updates to Local Proxy Mode with an already-installed upstream binary; otherwise perform a non-mutating availability check and leave installation to the explicit setup UI.

The existing review comments about untranslated French and Vietnamese values, the raw English skip reason, and the view-level update-button race also remain valid. The view-level race should be addressed together with the manager-wide serialization issue above.

Verification: the exact head 2323015e builds successfully in Debug, and a clean merge with the current master ec770e56 also builds successfully. git diff --check still reports three trailing-whitespace lines.

nguyenphutrong (commented):
Follow-up review against current `master` (`1d9e30d`) on the unchanged head `2323015e`:

The previously requested correctness changes remain unresolved: the selected binary source is still read again after suspension, proxy upgrades are not serialized across the Settings and menu entry points, and a missing upstream installation is still treated as needing an automatic install. As a result, the generic app update command can silently install CLIProxyAPI in Remote/Monitor mode, concurrent invocations can corrupt shared dry-run state, and changing the source during the release request can misclassify an upstream binary as the bundled track.

The current app already exposes a dedicated proxy update card under Settings/About with a non-mutating check followed by an explicit Update action. This PR adds a second proxy-update surface and couples Sparkle's generic “Check for Updates…” command to an implicit binary installation, which is both duplicative and less safe than the existing flow. The localization findings from the earlier review also remain unresolved, and there is no targeted coverage for source changes, concurrent invocations, or missing-install behavior.

The branch still merges cleanly; the synthetic merge builds successfully and the current test suite passes. Those tests do not exercise the update paths above. Given the unchanged implementation and the existing explicit update flow, I recommend closing this PR rather than merging it; any renewed combined-update behavior should start from the current architecture in a narrowly scoped PR.

nguyenphutrong:
Thank you for contributing this update workflow and for the effort put into the implementation. After re-reviewing it against the current codebase, we've decided not to move forward with this PR. Quotio now has an explicit proxy update flow in Settings, while coupling the generic app update command to an automatic proxy installation introduces source-switching and concurrent-update risks and may install a local proxy in modes where it is not needed. Since the branch has not been updated to address the outstanding review findings and the surrounding architecture has evolved, a fresh, narrowly scoped PR would be safer if this behavior is revisited. Thank you again for helping improve Quotio.

PR #420, opened 2026-06-09
Allow custom provider setup with optional SSL verification

### LINKED ISSUE #376: 新增自定义提供商报错
### Your Question

公司的API是HTTP，新增自定义提供商报错：The resource could not be loaded because the App Transport Security policy requires the use of a secure connection.
昨天还可以走HTTP的

### Topic

Custom Providers

### What have you tried?

_No response_

### Additional Context

_No response_

### Quotio Version (optional)

_No response_

jsyt:
## Summary
- Add per-custom-provider settings for trusted insecure endpoints.
- Keep HTTPS certificate verification enabled by default, with a warning before disabling it.
- Keep insecure HTTP blocked by default, with a separate warning before allowing it.
- Use the selected settings for Custom Provider model fetching and connection testing without globally relaxing ATS.
- Document the new transport and provider security fields for docstring coverage.

## Background
Custom Provider setup can fail in Quotio even when the same endpoint works from Claude Code or Codex CLI. There are two distinct macOS networking failures involved:

1. HTTPS endpoints can fail during URLSession trust evaluation when Apple TLS policy rejects a certificate that other CLI stacks accept. One concrete case from #376 is an internal endpoint whose certificate chain can be trusted manually but whose leaf certificate exceeds the maximum allowed validity period. That produces the app-level error: "A TLS error caused the secure connection to fail."

2. HTTP endpoints fail earlier. For `http://model.mify.ai.srv/v1`, URLSession is blocked by App Transport Security before any request is sent, producing: "The resource could not be loaded because the App Transport Security policy requires the use of a secure connection."

## Approach
This keeps secure behavior as the default and scopes the unsafe behavior to Custom Provider setup/test requests only.

- HTTPS still uses URLSession. If SSL verification is disabled for that provider, the setup/test session accepts the server trust challenge.
- HTTP does not use URLSession, because ATS blocks arbitrary `http://` URLs before request handling. If insecure HTTP is explicitly allowed for that provider, setup/test requests use a small plain HTTP transport for the model list request.

I intentionally did not set `NSAllowsArbitraryLoads`, because that would weaken every network request in the app. I also did not hard-code an ATS exception for `model.mify.ai.srv`, because custom provider hosts are user-entered at runtime and upstream should not carry a private host exception.

I also did not emit `verify-ssl` or `allow-insecure-http` into generated CLIProxyAPI YAML because this repository does not show a supported custom-provider YAML contract for those options.

## Testing
- `git diff --check`
- `node -e "JSON.parse(require('fs').readFileSync('Quotio/Localizable.xcstrings','utf8')); console.log('Localizable.xcstrings JSON OK')"`
- `swiftc -parse Quotio/Models/CustomProviderModels.swift Quotio/Services/CustomProviderService.swift Quotio/Services/CustomProviderHTTPClient.swift Quotio/Views/Components/CustomProviderSheet.swift`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swiftc -typecheck -swift-version 6 -default-isolation MainActor Quotio/Services/CustomProviderHTTPClient.swift`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS' build`

Xcode: 26.5 (build 17F42)

Fixes #376

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -4071,6 +4071,296 @@
         }
       }
     },
+    "customProviders.allowInsecureHTTP" : {
+      "extractionState" : "stale",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Allow Insecure HTTP"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Autoriser HTTP non sécurisé"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Cho phép HTTP không an toàn"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "允许不安全的HTTP"
+          }
+        }
+      }
+    },
+    "customProviders.allowInsecureHTTP.warning" : {
+      "extractionState" : "stale",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "HTTP is unencrypted. Requests to this provider may expose API keys and responses."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "HTTP n'est pas chiffré. Les requêtes vers ce fournisseur peuvent exposer les clés API et les réponses."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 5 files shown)


nguyenphutrong:
Thank you for the thoughtful contribution and the detailed testing provided here.

This pull request currently has merge conflicts with `master` and has had no activity since June 9, 2026. To keep the open pull request queue current and actionable, this PR is being closed.

If you would like to continue this work, please rebase it onto the latest `master`, resolve the conflicts, address any remaining review feedback, and feel free to reopen the PR or submit a new one. The contribution is appreciated.

PR #438, opened 2026-07-19
Make Monitor mode independent from CLIProxyAPI

nguyenphutrong:
## Summary

- run Monitor mode quota discovery and refresh independently from CLIProxyAPI
- preserve Codex account identities and deduplicate legacy aliases across the app and menu bar
- keep Monitor providers visible even when their CLI binaries are not installed
- isolate app-hosted unit tests from production bootstrap and Keychain access
- cover the monitor runtime behavior with regression tests

## Why

Monitor mode still had two hidden runtime dependencies: the menu bar filtered providers by installed CLI binaries, and the app-hosted test target launched the production bootstrap before tests could run. This PR removes both dependencies so Monitor mode and its tests operate independently as intended.

## Validation

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug -derivedDataPath /tmp/quotio-monitor-review-derived test` (15/15 passed)
- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug -derivedDataPath /tmp/quotio-monitor-review-derived build`


nguyenphutrong:
@codex review

PR #448, opened 2026-08-05
fix: purge stale/hidden quota tiles from Quota screen

owenmead:
## Summary
- Disabling a monitor account (e.g. "Claude Desktop") only removed it from the Accounts
  list; `providerQuotas` kept its last-fetched entry until the next refresh cycle, so the
  Quota screen kept showing a tile for an account the user had just disabled.
- Separately, `discoverAccounts()` intentionally hides generic placeholder accounts (e.g.
  "Claude Code") once a more specific credential exists for the same provider, but the
  underlying quota fetch keeps running for them regardless — with no account row to
  disable, that tile was effectively stuck on permanently.
- `setMonitorAccountDisabled` now purges the disabled account's entry from
  `providerQuotas` immediately (mirroring the existing `deleteMonitorAccount` pattern),
  and `removeDisabledMonitorQuotas` also drops quota data for any placeholder key that
  discovery has hidden in favor of a more specific account. The placeholder table is now
  shared via `MonitorRefreshCoordinator.placeholderAccountKeys` instead of duplicated
  between the two call sites.

## Test plan
- [x] Built and ran locally via `./scripts/build_and_run.sh`
- [x] Reproduced both stale tiles (disabled account still showing; unremovable "Claude
      Code" placeholder tile) on current `master`
- [x] Confirmed both tiles disappear after the fix, in both the main window Quota screen
      and the menu bar dropdown

### DIFF
--- a/Quotio/Services/Monitor/MonitorRuntime.swift
+++ b/Quotio/Services/Monitor/MonitorRuntime.swift
@@ -562,6 +562,20 @@ actor MonitorRefreshCoordinator {
     private var retryAfter: [AIProvider: Date] = [:]
     private(set) var issues: [AIProvider: MonitorRefreshIssue] = [:]
 
+    /// Generic account keys that represent the same underlying login as a more
+    /// specific, higher-priority credential for the same provider (e.g. a
+    /// nameless "Claude Code" native credential once "Claude Desktop" is also
+    /// discovered). `discoverAccounts` hides these from the account list to
+    /// avoid duplicate rows; `QuotaViewModel` uses the same table to keep quota
+    /// data for hidden placeholders from lingering on the Quota screen.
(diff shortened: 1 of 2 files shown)


nguyenphutrong (approved):
Reviewed against current master (1d9e30d8). The change keeps placeholder suppression aligned with account discovery, removes disabled quota data before persisting the snapshot, and applies hidden-placeholder cleanup before the normal menu/status refresh path. I found no blocking correctness or concurrency issue.

Validation:
- Debug build passed.
- The full test run had one failure in the pre-existing timing-sensitive `testCoordinatorCoalescesConcurrentRefreshForProvider`; that exact test passed on an immediate isolated rerun, and the PR does not change the refresh/coalescing implementation.

PR #449, opened 2026-08-05
feat: Improve UI/UX/Design of Card and Ring dashboard and menu popover

owenmead:
## Summary
Session/Weekly/Extra tiles showed the reset countdown as tiny tertiary caption
text, making it hard to tell at a glance which account is about to reset (and
therefore which one to switch to) versus one with plenty of quota left but a
long wait until reset.

**Card style**
- Reset countdown (or "unused" / usage fraction / "off") now renders at the
  same size and weight as the percentage, in its own column.
- The standalone percentage text was removed as redundant with the bar itself;
  the "hero" slot only falls back to showing it when there's no countdown or
  usage fraction available.
- Thicker bar, tighter row spacing, no per-row caption line.

**Ring style — main window**
- Rings are larger and now use a real SwiftUI `Grid` (not `LazyVGrid`) with a
  fixed minimum of 3 columns, so Session/Weekly/Extra line up identically
  across every account card even when one account is missing a metric.
  Previously, a 2-metric account (e.g. Session/Weekly only) would reflow into
  wider columns and visibly drift out of alignment with 3-metric accounts.
- The ring-embedded percentage is smaller, medium-weight, and colored by
  status instead of bold/primary.

**Ring style — menu bar popover**
- Same fixed-column-count alignment fix, applied to a flat row instead of a
  multi-row grid — the popover only shows the ring + countdown (no name label
  or window caption, since there's more room for that detail in the main
  window).
- Footer timestamp ("x minutes ago") is left-justified instead of right.

The "4h window" / "7d window" captions and the disabled-account "off" state
share a small lookup by known metric name; unrecognized metric names simply
omit the caption rather than guess.

## Test plan
- [x] Built and ran locally via `./scripts/build_and_run.sh`
- [x] Verified Card and Ring styles in both the main Quota screen and the menu
      bar popover, across accounts with 2 and 3 metrics
- [x] Verified alignment holds for an account missing a metric (e.g. no
      "Extra") next to accounts that have all three
- [x] Verified Used/Remaining display mode toggle still works with the new
      layouts

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -13937,6 +13937,151 @@
         }
       }
     },
+    "quota.state.off" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "off"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "désactivé"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "tắt"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "关闭"
+          }
+        }
+      }
+    },
+    "quota.state.unused" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "unused"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "inutilisé"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "chưa dùng"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "未使用"
+          }
+        }
+      }
+    },
+    "quota.state.notEnabled" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "not enabled"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "non activé"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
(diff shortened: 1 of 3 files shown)


nguyenphutrong (changes requested):
Thanks for improving the reset-time hierarchy. The branch builds and the full test suite passes, but I found several correctness/usability issues that need to be addressed before merge:

1. **The universal “4h window” caption is incorrect.** `RingSlotData.windowCaption` derives cadence only from the display name `Session`, but Claude maps its explicitly five-hour window to that same name (`five-hour-session`), while Codex classifies variable windows up to 24 hours as `Session`. Please derive captions from actual window metadata/raw model identity, or omit the duration rather than showing a wrong one.

2. **Unknown quotas are rendered as real values in the menu.** `ModelQuota` uses a negative percentage for unavailable/unknown data. `RingGridLayout` passes that through as `-1` remaining / `101` used, while `menuHeroText` clamps it into `0%` / `100%`; `CardGridLayout` has the same semantic problem. This should show an unavailable placeholder, not a measured percentage or “off” unless there is an explicit enabled-state signal.

3. **Unlimited metrics with usage are mislabeled “unused.”** The hero helpers check `remainingPercent >= 100` before `formattedUsage`. Cursor on-demand intentionally uses 100% when no limit exists while still carrying a non-zero used count, so values such as “123 used” become “unused” in card/ring/menu layouts. Prefer the explicit usage presentation for these metrics and reserve “unused” for a genuinely zero-used window.

4. **The ring alignment is positional rather than semantic.** Both layouts place `models[index]` into a fixed number of slots. If Session is absent but Weekly exists, Weekly moves into the first slot; separate `Grid` instances also do not share column sizing across account cards. Please map known metrics to stable semantic slots (with empty placeholders), and retain enough labeling/accessibility context to identify each ring—the menu ring currently removes metric names entirely, including for nonstandard/Antigravity groups.

Validation on current master (`1d9e30d8`): Debug build passed; full macOS test suite passed.

owenmead:
Reopened as #527 — GitHub wouldn't reopen this one after the branch history was replaced.

All four issues addressed: captions now derive from the raw metric identity and are omitted when it states no window; unknown quotas render an availability placeholder instead of a clamped percentage or "off"; unlimited metrics show their usage rather than "unused"; and rings sit in stable semantic columns with fixed widths and accessibility labels.

I adopted your `ringPercent`/`unknownPercent` mechanism and your `ModelBadgeData.usage` field rather than keeping this PR's parallel versions.

PR #465, opened 2026-08-10
fix: stop repeated keychain prompts for Claude Code credentials

owenmead:
## Problem

Quotio reads the Claude Code CLI's `Claude Code-credentials` keychain item on **every quota poll**. That item belongs to the CLI, not to Quotio, so macOS gates it behind an authorization prompt — and clicking "Always Allow" does not make the prompt stop. It reappears several times a day, indefinitely.

The cause is how the CLI persists refreshed tokens. From the `claude` binary's keychain module:

```
security find-generic-password -a "…" -w -s "…"
add-generic-password -U -a "…" -s "…" -X "…"
delete-generic-password              ← fallback path in the same module
```

`security add-generic-password -U` does not update in place. It rebuilds the item with a fresh default ACL containing only the creating process, discarding every trusted-application entry — including the one "Always Allow" added for Quotio. The grant therefore survives only until the CLI's next token refresh.

Compounding it, the keychain read happened *before* the quota cache was consulted, so a wiped ACL produced a prompt per poll rather than one.

## Fix

Copy the credential into Quotio's own keychain item once, then serve quota from that copy. Nothing reads the CLI's item again, so the prompt appears at most once per install.

`MonitorAccountDiscovery.discover()` performs the one-time adoption; the account then arrives via `vault.accounts()` instead of the per-pass probe that was in `discoverNativeKeychains`.

## Staying read-only toward Claude Code

Quotio is a usage monitor and shouldn't interfere with the CLI it observes. Being read-only locally is not sufficient on its own, because refreshing is itself a write — to Anthropic's authorization server, spending a refresh token the CLI also holds.

The CLI treats that token as a serialized, single-use resource:

```
tengu_oauth_token_refresh_lock_acquiring / _acquired / _retry / _retry_limit_reached
ELOCKED, oauth_refresh_lock_timeout, _lock_compromised_pre_post
tengu_oauth_token_refresh_race_resolved / _race_recovered
tengu_oauth_refresh_compromised_cas_adopted_sibling
tengu_oauth_refresh_token_marked_dead_invalid_grant
tengu_oauth_refresh_token_cleared_on_disk, known_dead_refresh_token
```

A cross-process lock, sibling-race detection, and a dead-token path on `invalid_grant`. An unsynchronized refresh from Quotio could push the CLI into that dead-token state and log the user out of Claude Code.

So the adopted credential **carries no refresh token at all** — it is dropped at parse time, making it structurally impossible for Quotio to spend rather than merely disallowed. Both refresh branches in `fetchOwnedQuota` are gated on `credential.refreshToken` and are inert for a copy.

`~/.claude/.credentials.json` had the same problem through a different door — it was being refreshed and rewritten in place — so it is now read without refresh or write-back. Auth files under `~/.cli-proxy-api/`, which Quotio and the proxy own, still refresh as before.

After the change: one read of the CLI's keychain item in the codebase (the one-time adoption), zero writes, and all four `refreshAccessToken` call sites unreachable for CLI-owned credentials.

## Trade-off

An adopted copy cannot be renewed once its access token expires, so it surfaces as needing authentication. Signing in to Quotio yields a credential with an independent refresh lineage that can be kept alive indefinitely and never touches Claude Code — that flow already existed (`beginClaudeLogin`/`completeClaudeLogin`) and is unchanged. `ownedAccount()` prefers a signed-in credential over an adopted copy when both exist.

## Testing

New `ClaudeCredentialAdoptionTests` covers the refresh-token drop, credential parsing (including millisecond expiry and the email fallback), adoption being skipped once a copy exists, copy-vs-login distinction, login precedence, and adopted accounts outranking native candidates for the same login.

Full suite: 73 passing.

One note for maintainers: `MonitorRuntimeTests.testCoordinatorCoalescesConcurrentRefreshForProvider` is a pre-existing load-sensitive flake, unrelated to this change — I verified it failing on unmodified `master` under CPU load (1 of 2 runs). `refresh(provider:)` only touches the `inFlight` map and never reaches adoption. Might be worth tightening separately.

### DIFF
--- a/Quotio/Services/Monitor/ClaudeCredentialAdoption.swift
+++ b/Quotio/Services/Monitor/ClaudeCredentialAdoption.swift
@@ -1,0 +1,132 @@
+import Foundation
+
+/// Copies the Claude Code CLI's keychain credential into Quotio's own keychain
+/// item so quota polling never has to read an item Quotio does not own.
+///
+/// The CLI persists refreshed tokens with `security add-generic-password -U`,
+/// which rebuilds the item from scratch and discards its ACL. Any "Always Allow"
+/// grant the user gives Quotio therefore survives only until the CLI's next
+/// token refresh, so reading that item on a poll loop produces an authorization
+/// prompt several times a day. Adopting the credential once trades all of those
+/// for a single prompt.
+///
+/// The copy is strictly read-only with respect to the CLI. Quotio never writes
+/// the CLI's item and never refreshes the adopted credential: the CLI serializes
+/// refreshes across its own processes with a lock and treats a spent refresh
+/// token as fatal (`invalid_grant` -> token marked dead -> cleared from disk),
+/// so an unsynchronized refresh from Quotio could log the user out of Claude
+/// Code. The adopted copy is used until it expires and no further; continuous
+/// monitoring comes from signing in to Quotio, which yields an independent
+/// credential Quotio may refresh freely.
+actor ClaudeCredentialAdopter {
+    static let shared = ClaudeCredentialAdopter()
+
+    nonisolated static let service = "Claude Code-credentials"
+
+    /// Marks a vault account as a copy of the CLI's credential rather than one
+    /// obtained by signing in to Quotio. Refreshing is only safe for the latter.
+    nonisolated static let credentialReference = "keychain:claude-code-copy"
+
+    /// Wait before touching the CLI's item again after an adoption that did not
+    /// yield a usable credential — a denied prompt must not become a prompt per
+    /// poll.
+    private let failureBackoff: TimeInterval = 6 * 3600
+
+    private let vault: MonitorCredentialStore
+    private var nextAttempt: Date?
+
+    init(vault: MonitorCredentialStore = MonitorCredentialVault.shared) {
+        self.vault = vault
+    }
+
+    /// True when the account is a copy of the CLI's credential, whose refresh
+    /// token is shared with Claude Code and must not be spent.
+    nonisolated static func isCopyOfCLICredential(_ account: MonitorAccount) -> Bool {
+        account.credentialReference == credentialReference
+    }
+
+    /// The Quotio-owned Claude credential, performing the one-time copy from the
+    /// CLI's keychain item if it has not happened yet.
+    @discardableResult
+    func adoptIfNeeded() async -> MonitorAccount? {
+        if let owned = await ownedAccount() { return owned }
+        return await adopt()
+    }
+
+    /// Existing Quotio-owned Claude credential, if any. Never reads the CLI item.
+    /// An account from signing in to Quotio takes precedence over an adopted copy,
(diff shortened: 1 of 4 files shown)


nguyenphutrong:
Thanks for the thorough investigation — especially the analysis around Claude Code’s single-use refresh-token lineage. There are useful safety improvements here, but this automatic adoption flow should not be merged as written after #466.

#466 now makes every background read/update of credentials owned by other applications strictly non-interactive (`kSecUseAuthenticationUIFail` plus process-level interaction suppression). This changes a core assumption of this PR: when Quotio is not already authorized to read `Claude Code-credentials`, `readExternalCredentialRecord` returns `nil` and cannot show the proposed one-time prompt. The adopter therefore cannot copy the credential for the users this PR is targeting.

The review also found three behavioral issues:

1. Having any Quotio-owned Claude credential prevents adoption/discovery of a different Claude Code account.
2. Disabling an adopted account does not stop it from being polled, because disabled state is stored separately from the raw vault account’s `isDisabled` property.
3. `fetchAdoptedNativeQuotas` can return a normal Quotio login even when `includeMonitorCredentials` is `false`.

The valuable part to preserve is the read-only behavior toward Claude Code: do not refresh or write back to `Claude Code-credentials`, and do not refresh or rewrite `~/.claude/.credentials.json`. Files under `~/.cli-proxy-api/` can continue refreshing as before.

Would you be willing to rebase onto current `master` and reduce this PR to those read-only changes, removing the background adopter? If credential import is added later, it should be an explicit user-initiated action while all discovery and polling remain non-interactive.

I tested the effective merge with current `master`: all 74 tests pass and the Debug build succeeds, but the current tests do not cover the interaction, disabled-state, multi-account, or `includeMonitorCredentials` boundaries above.

nguyenphutrong (changes requested):
Requesting changes because the automatic adoption path is incompatible with the non-interactive external-keychain contract now on `master`: when Quotio lacks access, the promised one-time prompt cannot occur and no credential is copied. The adopter also ignores persisted disabled state, conflates a Quotio login with an adopted CLI account, and bypasses the `includeMonitorCredentials` boundary. Please see the detailed review comment for the affected flows and suggested narrower scope. The effective merge builds successfully and all 74 tests pass, but those tests do not cover these behavioral boundaries.

owenmead:
Reopened as #526 — GitHub wouldn't reopen this one after the branch history was replaced.

Reduced to the read-only scope you asked for: the background adopter is gone (which removes all three behavioural issues along with the #466 incompatibility), leaving only the no-refresh/no-write-back boundary toward `Claude Code-credentials` and `~/.claude/.credentials.json`. `~/.cli-proxy-api/` still refreshes.

Also worth flagging: #466 already fixed the repeated-prompt symptom this PR was named for, so #526 is retitled and rescoped to what it actually does — not spending the CLI's single-use refresh token.

Issue #367, opened 2026-03-18
[Bug] Codex CLI session history disappears when proxy is configured — auth.json overwrites account_id


zhd4n:
## Summary
When Quotio configures Codex CLI via "Automatic" mode, it fully overwrites 
`~/.codex/auth.json` with a proxy key, removing the native `account_id` field. 
This causes `codex resume` to show no previous sessions — they become invisible 
even though the JSONL files still exist on disk.

## Environment
- Quotio: latest
- Codex CLI: 0.111.0
- macOS (arm64)

## Steps to reproduce
1. Use Codex CLI natively for some time — accumulate session history
2. Open Quotio → Agents → Codex CLI → Configure → Automatic
3. Launch `codex resume` — session list is empty

## Root cause
Before configuration, `~/.codex/auth.json` contains:
```json
{
  "access_token": "eyJ...",
  "account_id": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
}
```

After Quotio configures Codex, `auth.json` becomes:
```json
{
  "OPENAI_API_KEY": "quotio-local-UUID"
}
```

Codex CLI filters sessions by `account_id` from `auth.json`. 
Without it, `codex resume` returns no results.

Note: Quotio correctly backs up `config.toml` 
(as `config.toml.backup.{timestamp}`), but does **not** back up `auth.json`.

## Expected behavior
Quotio should either:
- **Preserve** existing fields in `auth.json` and only add/update `OPENAI_API_KEY`, or
- **Back up** `auth.json` before overwriting (same as it does for `config.toml`)

## Workaround
Manually save native `auth.json` before configuring:
```bash
cp ~/.codex/auth.json ~/.codex/auth.json.native
# restore when needed:
cp ~/.codex/auth.json.native ~/.codex/auth.json
```
The native `auth.json` works fine even when the proxy is active — 
CLIProxyAPI does not actually validate the proxy key from Codex side.

PR #467, opened 2026-08-12
fix(codex): preserve auth.json credentials and back the file up when configuring proxy

giaBaoJS:
## Scope

Related issue: #367

It keeps the `auth.json` data-preservation and backup work, and drops the incorrect `account_id` root-cause claim. #367's actual symptom — the empty `codex resume` list — is caused by the `model_provider` switch and is **not** fixed here; findings and a suggested direction are in the review thread.

## What this fixes

**1. Configuring the proxy destroyed the user's Codex login.**
`generateCodexConfig` rebuilt `~/.codex/auth.json` from scratch as `{"OPENAI_API_KEY": "quotio-local-..."}` and, in Automatic mode, wrote it over the existing file. Codex CLI stores its own credentials there (`tokens.access_token`, `tokens.refresh_token`, `account_id`, `last_refresh`), and unlike `config.toml` the file was never backed up — so a ChatGPT-login user was signed out with no recovery path.

`mergedCodexAuthJSON(existing:apiKey:)` now reads the existing file, preserves every field, and only sets `OPENAI_API_KEY`. A corrupt file degrades to managed-only content with a warning, matching the existing `config.toml` merge fallback.

**2. `auth.json` is now backed up.** Automatic configure backs it up via the collision-safe `backupIfPresent` helper (which never overwrites an existing backup); the `config.toml` backup was switched to the same helper. `CLIAgent.codexCLI.configPaths` already lists `~/.codex/auth.json`, so these appear in `listBackups` and restore through the existing flow.

**3. Manual mode no longer exposes native credentials.** The merged content is only ever written to disk in Automatic mode. `codexAuthPayloads(existing:apiKey:path:)` returns two renderings — `managed` (the Quotio key alone) for the manual preview, `merged` for the automatic write — so access/refresh tokens never reach `RawConfigOutput`, the preview UI, or "Copy All". The preview instruction now tells the user to *add* the key rather than replace the file.

**4. "Use default" cleans both files independently.** The `auth.json` cleanup was nested inside `if fileExists(config.toml)`, so a user whose `config.toml` was already gone kept the Quotio-managed key forever while the revert reported success. The two files are now handled separately, both with collision-safe backups. Only a `quotio-`-prefixed value is removed, so a user's own `OPENAI_API_KEY` is never touched; permissions stay `0600`.

## Why merging rather than backup-only

Proxy routing is decided by `config.toml` (`model_providers.cliproxyapi.base_url`), not by `auth.json`, and as noted in #367 CLIProxyAPI does not validate the key Codex sends. Keeping the native token fields alongside `OPENAI_API_KEY` therefore cannot redirect traffic away from the proxy. Backups are kept as a second recovery path.

## Test evidence

`QuotioTests/CodexAuthConfigTests` — 15 tests:

- merge preserves `tokens`/`account_id`/`last_refresh` and updates only `OPENAI_API_KEY`; no existing file yields managed-only; corrupt input throws.
- `testManualPayloadNeverExposesNativeCredentials` — asserts the manual preview string contains none of the native token values (regression test for 3).
- `testAutomaticPayloadMergesIntoExistingFile`, `testPayloadsFallBackToManagedOnlyWhenExistingFileIsCorrupt`.
- `testRevertRemovesManagedKeyWithoutAnyConfigTOML` — reverts with **no `config.toml` present at all** (regression test for 4), checking the key is gone, native fields survive, a backup exists, and permissions are `0600`.
- `testRevertLeavesUserOwnedKeyAndReportsNoChange`, `testRevertIsANoOpWhenAuthFileIsMissing`.
- `testRepeatedRevertsKeepEveryBackup` — two reverts in the same second keep both backups.
- `testRestoreFromBackupRestoresOriginalAuthJSON` — restore returns the original bytes.

```
xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build   # BUILD SUCCEEDED
xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'   # TEST SUCCEEDED — 80 tests, 0 failures
```

New user-facing strings are localized in en, fr, vi and zh-Hans.

Trial-merged against `master` and against #473, #477, #482 and #484 in a scratch worktree — no conflicts in any combination.

Refs #367

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -1422,6 +1422,35 @@
         }
       }
     },
+    "agents.codex.authJSONMergeKey" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Add this key to ~/.codex/auth.json, keeping any entries already in the file:"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Ajoutez cette clé à ~/.codex/auth.json en conservant les entrées déjà présentes dans le fichier :"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Thêm khóa này vào ~/.codex/auth.json, giữ nguyên các mục đã có trong tệp:"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "将此密钥添加到 ~/.codex/auth.json，并保留文件中已有的条目："
+          }
+        }
+      }
+    },
     "agents.codex.mergeAndSaveFiles" : {
       "extractionState" : "stale",
       "localizations" : {
@@ -1447,6 +1476,35 @@
           "stringUnit" : {
             "state" : "translated",
             "value" : "合并并将以下文件保存到 ~/.codex/ 目录："
+          }
+        }
+      }
+    },
+    "agents.codex.revertManualInstructions" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Remove the [model_providers.cliproxyapi] section from ~/.codex/config.toml, and remove the Quotio-managed OPENAI_API_KEY entry from ~/.codex/auth.json while keeping the other entries in that file."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Supprimez la section [model_providers.cliproxyapi] de ~/.codex/config.toml, puis supprimez l'entrée OPENAI_API_KEY gérée par Quotio de ~/.codex/auth.json en conservant les autres entrées de ce fichier."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Xóa phần [model_providers.cliproxyapi] khỏi ~/.codex/config.toml, và xóa mục OPENAI_API_KEY do Quotio quản lý khỏi ~/.codex/auth.json trong khi vẫn giữ các mục khác trong tệp đó."
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "从 ~/.codex/config.toml 中移除 [model_providers.cliproxyapi] 部分，并从 ~/.codex/auth.json 中移除由 Quotio 管理的 OPENAI_API_KEY 条目，同时保留该文件中的其他条目。"
           }
         }
       }
--- a/Quotio/Services/AgentConfigurationService.swift
+++ b/Quotio/Services/AgentConfigurationService.swift
@@ -655,15 +655,16 @@ actor AgentConfigurationService {
     private func generateCodexDefaultConfig(mode: ConfigurationMode) -> AgentConfigResult {
         let home = fileManager.homeDirectoryForCurrentUser.path
         let configPath = "\(home)/.codex/config.toml"
-        
+        let authPath = "\(home)/.codex/auth.json"
+
         if mode == .automatic && fileManager.fileExists(atPath: configPath) {
             do {
                 let content = try String(contentsOfFile: configPath, encoding: .utf8)
-                
+
                 // Create backup
                 let backupPath = "\(configPath).backup.\(Int(Date().timeIntervalSince1970))"
                 try content.write(toFile: backupPath, atomically: true, encoding: .utf8)
-                
+
                 // Reuse the same TOML-aware filtering used by merge path,
                 // including managed banner removal.
                 let stubManagedConfig = buildManagedCodexTOML(model: "", proxyURL: "")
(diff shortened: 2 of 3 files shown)


nguyenphutrong (changes requested):
Thanks for addressing the destructive `auth.json` overwrite. Preserving the file and adding collision-safe backups are valuable, but I found two blockers and one incomplete behavior:

1. **This does not fix the reported empty `codex resume` list.** The root-cause comment at lines 1021–1024 is incorrect for the issue's Codex 0.111.0. That release's resume picker filters rollouts by the current model provider and cwd, not by `account_id` ([resume picker](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/tui/src/resume_picker.rs#L119-L144), [thread list filters](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/core/src/rollout/list.rs#L105-L117)). Quotio still switches `model_provider` to `cliproxyapi`, so sessions recorded under the native provider remain hidden. The new tests verify JSON preservation but not the issue's observable behavior. Please reproduce the original flow and either solve the provider-filter boundary or narrow this PR so it no longer claims to fix #367.

2. **Manual mode now exposes native credentials.** `existingAuthData` is merged for every mode and then placed in `RawConfigOutput`; the manual preview renders that content and supports Copy All. A normal Codex file contains access/refresh tokens, so this puts secrets into the UI and clipboard. Please keep the merged write confined to automatic mode and make manual output an instruction/patch that does not include existing credentials.

3. **Default cleanup is incomplete.** Removing Quotio's key from `auth.json` is nested under `fileExists(config.toml)`. If `config.toml` is missing but `auth.json` remains, Automatic “Use default” reports success without removing the managed key. Please handle the two files independently (and keep using collision-safe backups).

The effective merge with current `master` builds successfully and all 78 tests pass. No standards violations were found; the blockers are behavioral/security boundaries that the tests currently do not cover.

giaBaoJS:
You're right, and I reproduced it. My `account_id` root cause was wrong — I've dropped that claim and the `Fixes #367` line, and re-scoped the PR to what it actually does.

## What I verified

`rust-v0.111.0` is an annotated tag; it dereferences to exactly the commit you linked:

```
$ gh api repos/openai/codex/git/ref/tags/rust-v0.111.0   --jq .object.sha
36ab739252724325262726dff8fbbcb26c2447ad          # tag object
$ gh api repos/openai/codex/git/tags/36ab7392... --jq '.object.sha, .tag'
8c75cd9afcd405d134530e53c78e5e0e4e5312a3
rust-v0.111.0
```

So this is the issue's release, not a later one. In that tree:

- `account_id` appears **nowhere** in `tui/src/resume_picker.rs` or `core/src/rollout/list.rs`.
- [`resume_picker.rs#L147-L171`](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/tui/src/resume_picker.rs#L147-L171) sets `default_provider = config.model_provider_id` and passes `provider_filter = vec![default_provider]` into `list_threads`. `show_all` only clears `filter_cwd` (L149-153) — `--all` does **not** relax the provider filter.
- [`list.rs#L975-L1000`](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/core/src/rollout/list.rs#L975-L1000) `ProviderMatcher::matches` keeps a rollout only if its recorded `model_provider` is in the filter list; rollouts with no recorded provider fall back to `matches_default_provider`.
- [`recorder.rs#L408`](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/core/src/rollout/recorder.rs#L408) stamps `model_provider: Some(config.model_provider_id.clone())` into `SessionMeta` at session creation.

So a session recorded under `openai` is filtered out the moment `model_provider = "cliproxyapi"` is active. Exactly as you said.

## Reproduction

Real `codex-cli 0.111.0` (npm `@openai/codex@0.111.0`), a scratch `CODEX_HOME` with one rollout recorded under `model_provider: "openai"`, `source: "cli"`. Driven with `codex exec resume --last`, which applies the same provider filter as the picker ([`tui/src/lib.rs#L778-L790`](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/tui/src/lib.rs#L778-L790)) but is scriptable. Native session id `019d8f0f-…-751111`; sessions dir reset between cases.

| case | `config.toml` | `auth.json` | `thread.started` |
|---|---|---|---|
| A | native (`openai`) | native | `019d8f0f-…-751111` — **found** |
| B1 | Quotio (`cliproxyapi`) | overwritten (pre-PR) | fresh thread — **hidden** |
| B2 | Quotio (`cliproxyapi`) | merged, `account_id` intact (this PR) | fresh thread — **hidden** |
| C | native (`openai`) | overwritten, `account_id` gone | `019d8f0f-…-751111` — **found** |

**B2 is the decisive one**: this PR's merge does not bring the session back. **C** is the converse: destroying `account_id` entirely does not hide anything as long as the provider is unchanged. `auth.json` is simply not an input to session listing in 0.111.0.

## Why I did not try to solve the provider boundary here

The obvious workaround would be to keep `model_provider_id == "openai"` and point the built-in provider at the proxy. That cannot work: [`config/mod.rs#L1855-L1858`](https://github.com/openai/codex/blob/8c75cd9afcd405d134530e53c78e5e0e4e5312a3/codex-rs/core/src/config/mod.rs#L1855-L1858) merges user providers with `model_providers.entry(key).or_insert(provider)` — **built-ins win**, so a user-defined `[model_providers.openai]` is silently ignored.

The remaining options are rewriting `model_provider` inside historical rollout JSONL files, or shipping a Quotio-side session browser. Both fight Codex's intended filtering, and on current `main` the filter is still there with `ProviderFilter::Any` reserved for remote workspaces only (`resume_picker.rs`, `picker_provider_filter`) — there's no supported opt-out. That's a product decision for you, not something to slip into a bugfix PR, so I've left #367 open with these findings rather than inventing a workaround.

## What's in the PR now

Blockers 2 and 3 are addressed:

- **Manual mode secrets** — merged content is now only produced for the Automatic write. `codexAuthPayloads` returns `managed` (Quotio key alone) for the preview and `merged` for disk, so tokens never reach `RawConfigOutput` / the preview / Copy All. The preview instruction now says *add this key*, not *save this file*. Regression test asserts the preview string contains none of the native token values.
- **Revert cleanup** — `config.toml` and `auth.json` are handled independently, both via the collision-safe `backupIfPresent`. Regression test reverts with no `config.toml` present at all.

The `auth.json` overwrite is still worth fixing on its own merits — it signs a ChatGPT-login user out with no backup — which is why I kept that part rather than closing the PR. Re-verified after the re-scope: 80 tests pass, build clean, and the branch trial-merges without conflict against `master` and against #473, #477, #482 and #484.

nguyenphutrong (changes requested):
The code-level blockers are resolved: native credentials stay out of manual preview, automatic writes preserve them, and auth cleanup is independent of config.toml. One blocking metadata issue remains: GitHub still reports #367 in this PR’s closingIssuesReferences, even though the revised scope correctly says the resume-provider bug is not fixed. The sentence “no longer claims to fix #367” still matches GitHub’s closing-keyword parser. Please reword/remove that phrase so merging cannot close #367; keep a non-closing reference such as “Does not resolve issue 367” or “Refs #367.”

PR #469, opened 2026-08-12
fix(menubar): deduplicate GitHub Copilot quota items

### LINKED ISSUE #404: [Bug] Duplicate GitHub Copilot menu bar entries from same auth file
### What happened?

A single GitHub Copilot auth file (`~/.cli-proxy-api/github-copilot-marcosvrs.json`) produces **two duplicate entries** in the menu bar:

1. `{"provider":"github-copilot","accountKey":"github-copilot-marcosvrs.json"}` — keyed by full filename
2. `{"provider":"github-copilot","accountKey":"marcosvrs"}` — keyed by `username` field or extracted suffix

Both entries point to the same file and the same account, consuming 2 of the `menuBarMaxItems` slots.

### Steps to reproduce

1. Create a Copilot auth file at `~/.cli-proxy-api/github-copilot-marcosvrs.json`:
   ```json
   {
     "access_token": "ghu_...",
     "token_type": "Bearer",
     "username": "marcosvrs",
     "type": "copilot"
   }
   ```
2. Start Quotio v0.17.0 in Quota-Only Mode
3. Observe **two** GitHub Copilot entries in the menu bar

### Expected behavior

One entry per auth file. The quota fetcher should use a single canonical key (e.g., the `username` field from JSON, or the extracted suffix from the filename) and deduplicate.

### Workaround attempts

- Patching `menuBarSelectedQuotaItems` in `dev.quotio.desktop.plist` → Quotio overwrites on next launch
- Setting `hasUserModifiedMenuBar: true` → still auto-adds duplicate
- Removing `username` from JSON → Quotio still creates two entries (filename-based + suffix-extracted)
- Sending SIGHUP to reload config → Quotio exits

### Root cause (suspected)

Two internal code paths produce different keys for the same file:
- `DirectAuthFileService` scan: uses full filename as `accountKey`
- `CopilotQuotaFetcher`: extracts suffix from filename or reads `username` field

### Related issues

- #134 — Similar keying mismatch for Copilot (email vs username)
- #382 — Codex usage not shown in menu bar

### Environment

- macOS 26.x Tahoe
- Quotio v0.17.0
- App Mode: Quota-Only Mode (monitor)
- Copilot plan: Business

giaBaoJS:
## Root cause

A single Copilot auth file (`~/.cli-proxy-api/github-copilot-marcosvrs.json`) could surface **two** menu bar entries keyed differently for the same account:

1. **Legacy keys survive forever.** Older Quotio versions keyed Copilot quota items by the full auth filename (`github-copilot-marcosvrs.json`). Those keys persist in the Monitor snapshot (`snapshots-v1.json`) and in `menuBarSelectedQuotaItems`, and `MonitorRefreshCoordinator.refresh` merges `previous` into every fresh result — so the stale filename key is never dropped even though the fetcher now produces the canonical `marcosvrs` key. Both keys then count as valid quota entries: prune keeps both, auto-select re-adds both, and the account list shows a duplicate quota-derived account. The mismatched count also permanently trips the "partial refresh" path (`fresh.count < previous.count`).
2. **Two discovery paths can still disagree on the key.** `DirectAuthFileService` read the GitHub handle from the JSON `login` field, but CLIProxyAPI Copilot auth files store it as `username` (the field `CopilotQuotaFetcher` uses). When the handle differs from the filename suffix, the scan and the fetcher generate different keys for the same file.

`migrateMenuBarSelections` already migrated Codex filename aliases (#435), but for Copilot it only registered `email` as an alias, so filename-keyed selections were never migrated.

## Fix

Mirrors the approach the repo already merged for Codex (#435 "deduplicate Codex quota items" and #444 "refresh matching legacy quota aliases"):

- **`CopilotQuotaFetcher.reconcileLegacyAliases(in:)`** (new, modeled on `CodexCLIQuotaFetcher.reconcileLegacyAliases`): collapses legacy quota keys (full filename, filename without `.json`, filename suffix) onto the canonical account key (`username` from JSON, filename suffix otherwise), keeping the freshest quota. Collapsing only happens when the alias maps unambiguously to exactly one canonical key and is not itself another account's canonical key. Applied in `refreshQuotasDirectly` right where the Codex reconciliation already runs, so the persisted snapshot self-heals on the next refresh.
- **`migrateMenuBarSelections`**: registers the filename-based aliases (full filename, filename without `.json`, filename suffix) for Copilot in both the direct-auth-file and proxy-auth-file loops — same alias set Codex uses — so existing `menuBarSelectedQuotaItems` plist entries under either legacy key migrate to the canonical key and deduplicate.
- **`DirectAuthFileService`**: reads the GitHub handle from `login` **or** `username`, aligning `menuBarAccountKey` with the key `CopilotQuotaFetcher` generates.
- **`AppBootstrap.resolveQuotaData`**: adds the Copilot filename-key fallback next to the existing `codexFilenameKey` fallback, so a not-yet-migrated legacy selection still resolves to the canonical quota entry.

## Tests

New `QuotioTests/CopilotQuotaFetcherTests.swift` (10 tests) reproducing the issue scenario:

- one auth file with a `username` field and quota entries under all three legacy keys reconciles to exactly one canonical entry with the freshest data;
- legacy filename key migrates even when `username` differs from the filename suffix;
- newer canonical data is preserved / newer legacy data is promoted;
- an alias colliding with another account's canonical key, or an ambiguous alias, is left untouched;
- `DirectAuthFile.menuBarAccountKey` resolves to the GitHub handle (or filename suffix) for `github-copilot-*.json` files.

Verification:

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED** (all suites, including the 10 new tests)

Fixes #404

### DIFF
--- a/Quotio/Models/Models.swift
+++ b/Quotio/Models/Models.swift
@@ -386,13 +386,11 @@ nonisolated struct AuthFile: Codable, Identifiable, Hashable, Sendable {
             // so filename-based keys keep same-email Plus/Team accounts distinct.
             return name.codexFilenameKey
         }
-        if providerType == .copilot {
-            if let account, !account.isEmpty {
-                return account
-            }
-            if let filenameKey = name.copilotFilenameKey {
-                return filenameKey
-            }
+        if providerType == .copilot,
+           let key = CopilotQuotaFetcher.canonicalAccountKey(filename: name, username: account) {
+            // Same canonical rule as CopilotQuotaFetcher / DirectAuthFileService (#404):
+            // identity field first, `github-copilot-*` filename suffix otherwise.
+            return key
         }
         if let email = email, !email.isEmpty {
             return email
--- a/Quotio/QuotioApp.swift
+++ b/Quotio/QuotioApp.swift
@@ -154,8 +154,13 @@ final class AppBootstrap {
             return quotaData
         }
 
-        guard provider == .codex else { return nil }
-        return accountQuotas[selectedItem.accountKey.codexFilenameKey]
+        if provider == .codex {
+            return accountQuotas[selectedItem.accountKey.codexFilenameKey]
+        }
+        if provider == .copilot, let filenameKey = selectedItem.accountKey.copilotFilenameKey {
+            return accountQuotas[filenameKey]
+        }
+        return nil
     }
 }
 
--- a/Quotio/Services/DirectAuthFileService.swift
+++ b/Quotio/Services/DirectAuthFileService.swift
@@ -21,7 +21,11 @@ struct DirectAuthFile: Identifiable, Sendable, Hashable {
     let filePath: String
     let source: AuthFileSource
     let filename: String
-    
+    /// Identity fields this file also matches, but which are not the canonical
+    /// key. Older Quotio versions keyed Copilot files by the JSON `login` field;
+    /// those keys must still migrate onto the canonical key (#404).
+    nonisolated var legacyIdentityKeys: [String] = []
+
     /// Source location of the auth file
     enum AuthFileSource: String, Sendable {
         case cliProxyApi = "~/.cli-proxy-api"
@@ -57,13 +61,9 @@ struct DirectAuthFile: Identifiable, Sendable, Hashable {
         if provider == .codex {
             return filename.codexFilenameKey
         }
-        if provider == .copilot {
-            if let login, !login.isEmpty {
-                return login
-            }
-            if let filenameKey = filename.copilotFilenameKey {
-                return filenameKey
-            }
+        if provider == .copilot,
+           let key = CopilotQuotaFetcher.canonicalAccountKey(filename: filename, username: login) {
+            return key
         }
         if provider == .kiro {
             if let email = email, !email.isEmpty {
@@ -172,9 +172,25 @@ actor DirectAuthFileService {
         
         // Extract metadata
         var email = json["email"] as? String
-        // Copilot auth files written by CLIProxyAPI store the GitHub handle as
-        // "username"; older files may use "login".
-        let login = json["login"] as? String ?? json["username"] as? String
+        var login = json["login"] as? String
+        var legacyIdentityKeys: [String] = []
+        if provider == .copilot {
+            // Copilot auth files written by CLIProxyAPI store the GitHub handle as
+            // "username"; older files use "login". The precedence must match
+            // CopilotQuotaFetcher, otherwise the direct-auth item and the quota
+            // result key the same file differently and #404 comes back.
+            let rawLogin = json["login"] as? String
+            login = CopilotQuotaFetcher.canonicalIdentityField(
+                username: json["username"] as? String,
+                login: rawLogin
+            )
+            // Releases before this fix keyed the file by `login`; keep that value
+            // as a migration alias when `username` now wins.
+            if let rawLogin = rawLogin?.trimmingCharacters(in: .whitespacesAndNewlines),
+               !rawLogin.isEmpty, rawLogin != login {
+                legacyIdentityKeys.append(rawLogin)
+            }
+        }
         let accountType = json["account_type"] as? String
         
         // For Kiro: if email is empty, try to use provider (e.g., "Google") as identifier
@@ -201,7 +217,8 @@ actor DirectAuthFileService {
             accountType: accountType,
             filePath: filePath,
             source: .cliProxyApi,
-            filename: filename
+            filename: filename,
+            legacyIdentityKeys: legacyIdentityKeys
         )
     }
     
--- a/Quotio/Services/DirectAuthFileService.swift
+++ b/Quotio/Services/DirectAuthFileService.swift
@@ -172,7 +172,9 @@ actor DirectAuthFileService {
         
         // Extract metadata
         var email = json["email"] as? String
-        let login = json["login"] as? String
+        // Copilot auth files written by CLIProxyAPI store the GitHub handle as
+        // "username"; older files may use "login".
+        let login = json["login"] as? String ?? json["username"] as? String
         let accountType = json["account_type"] as? String
         
         // For Kiro: if email is empty, try to use provider (e.g., "Google") as identifier
--- a/Quotio/Services/QuotaFetchers/CopilotQuotaFetcher.swift
+++ b/Quotio/Services/QuotaFetchers/CopilotQuotaFetcher.swift
@@ -148,6 +148,8 @@ nonisolated struct CopilotAuthFile: Codable, Sendable {
     let tokenType: String?
     let scope: String?
     let username: String?
+    /// Older CLIProxyAPI releases stored the GitHub handle under `login`.
+    let login: String?
     let type: String?
 
     enum CodingKeys: String, CodingKey {
@@ -155,6 +157,7 @@ nonisolated struct CopilotAuthFile: Codable, Sendable {
         case tokenType = "token_type"
         case scope
         case username
+        case login
         case type
     }
 }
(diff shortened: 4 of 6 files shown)


nguyenphutrong (changes requested):
Thanks for the thorough root-cause analysis and the good alias-collision coverage. The synthetic merge into the current `master` builds successfully and all 74 tests pass. I found two remaining paths that can preserve or recreate the duplicate, so this is not quite merge-ready yet:

1. Please reconcile the result of the provider-scoped Monitor refresh as well. In `QuotaViewModel.refreshMonitorProvider`, the `.copilot` case currently assigns the raw result of `coordinatedRefresh`, unlike the `.codex` case immediately above it. `MonitorRefreshCoordinator.refresh` merges `previous` into `fresh`, so when a user refreshes only Copilot, the stale filename key survives and is stored again. Applying `fetcher.reconcileLegacyAliases(in: refreshed)` there (and covering that integration/path) makes the fix consistent across full and scoped refreshes.

2. Please make `DirectAuthFileService` use the same Copilot canonical-field precedence as `CopilotQuotaFetcher`. The new parser expression prefers `login` whenever both `login` and `username` exist, while `CopilotQuotaFetcher` keys the same file by `username`. If those fields differ, the direct-auth item is keyed by `login` and the quota result by `username`, recreating two entries for one file. For Copilot, prefer `username` and fall back to `login`; keep the existing behavior for other providers. A regression test with both fields present and different would protect this contract.

Once these two identity paths agree, the rest of the reconciliation and migration approach looks appropriate for #404.

giaBaoJS:
Both points addressed in `71b3dfc`. I also audited every place a Copilot account key is derived or stored rather than only the two you named, since the previous round's miss was exactly that class of gap.

### Key-derivation inventory

| # | Site | Rule now used | Status |
|---|---|---|---|
| 1 | `CopilotQuotaFetcher.fetchAllCopilotQuotas` (`CopilotQuotaFetcher.swift:339`) | `quotaKey(filename:authFile:)` → `canonicalAccountKey` | routed through shared rule |
| 2 | `CopilotQuotaFetcher.fetchQuota(accountKey:)` (`:263`) | same | routed through shared rule |
| 3 | `CopilotQuotaFetcher.accountIdentity` / `readAccountIdentities` (`:597`, `:611`) | same | routed through shared rule |
| 4 | `DirectAuthFileService.parseAuthFileJSON` (`DirectAuthFileService.swift:174`) | `canonicalIdentityField(username:login:)` | **fixed** — was `login`-first |
| 5 | `DirectAuthFile.menuBarAccountKey` (`:64`) | `canonicalAccountKey(filename:username:)` | routed through shared rule |
| 6 | `AuthFile.quotaLookupKey` (management API, `Models.swift:389`) | same | routed through shared rule |
| 7 | `MonitorAccount.makeLegacy` (`MonitorRuntime.swift:74`) | delegates to `file.menuBarAccountKey` | covered via #5 |
| 8 | `QuotaViewModel.migrateMenuBarSelections` (`:2420`) | `file.menuBarAccountKey` + alias list | covered via #5, alias list extended |
| 9 | `AppBootstrap.resolveQuota` (`QuotioApp.swift:160`) | `accountKey.copilotFilenameKey` fallback | unchanged, already in this PR |
| 10 | `MonitorOAuthCoordinator.saveGitHubCredential` (`:314`) | GitHub API `login` (the handle) | already equals CLIProxyAPI `username`; no file involved |
| 11 | Native-token path in `fetchAllCopilotQuotas` (`:322`) | `fetchGitHubLogin` → same handle | no file involved |

### Refresh / store paths

| Path | Reconciled? |
|---|---|
| Full monitor refresh — `refreshQuotasDirectly` (`QuotaViewModel.swift:738`) | yes (previous round) |
| **Provider-scoped monitor refresh — `refreshMonitorProvider` `.copilot`** | **yes — fixed** |
| Per-account refresh — `refreshQuota(for account:)` (`:1851`) | n/a: writes a single entry at the caller-supplied canonical key, via `fetchQuota(accountKey:)` which now matches with the shared rule (#2) |
| Non-monitor refresh — `refreshCopilotQuotasInternal` (`:1728`), used by `refreshQuotasUnified` and non-monitor scoped refresh | n/a: `providerQuotas[.copilot] = quotas` is a full replacement, not a merge, and the fetcher only emits canonical keys, so no stale key can survive |
| Snapshot bootstrap — `MonitorRefreshCoordinator.bootstrap` (`:411`) | replays whatever was on disk; `initializeQuotaOnlyMode` immediately runs `refreshQuotasDirectly`, and `finish(quotas:)` (`:757` / `:1996`) rewrites the snapshot from the reconciled dictionary, so the stale keys are gone after the first refresh |

### 1. Provider-scoped refresh

You were right that the merge is the problem, and reconciling only the fetcher output would not have been enough either — `MonitorRefreshCoordinator.refresh` returns `previous.merge(fresh)`, so the reconciliation has to run on the merged dictionary. That is now `QuotaViewModel.refreshCopilotThroughCoordinator`, which owns both steps so the scoped path cannot drift from the full one again.

New test `testScopedMonitorRefreshCollapsesStaleKeyFromCoordinatorMerge` drives the real `MonitorRefreshCoordinator` and the real production helper against a temp auth dir. It asserts the counterfactual inline: the coordinator merge alone yields `{"github-copilot-marcosvrs.json", "marcosvrs"}`, and the production path yields `{"marcosvrs"}` with the fresh timestamp.

Proof it fails without the fix — replacing the helper body with `return merged`:

```
Test case '...testScopedMonitorRefreshCollapsesStaleKeyFromCoordinatorMerge()' failed (1.908 seconds)
```

with every other test in the file still passing. Restored afterwards.

### 2. Canonical-field precedence

Fixed to `username`, then `login`, as you asked. Rather than duplicating the precedence, it is now one function, `CopilotQuotaFetcher.canonicalIdentityField(username:login:)`, with `canonicalAccountKey(filename:username:login:)` adding the filename-suffix fallback — called from all of sites 1-6 above. A duplicated rule is what caused this, so there is only one now.

I also made `CopilotAuthFile` decode `login`. Without it the mismatch persisted in the other direction: on a legacy file that has only `login`, the parser keyed by `login` while the fetcher, which never decoded that field, fell back to the filename suffix.

New test `testParsedAuthFileAndFetcherAgreeWhenLoginAndUsernameDiffer` writes `github-copilot-work.json` with `username: marcosvrs` / `login: marcos-legacy`, runs the real `DirectAuthFileService` scan, and asserts the parser and `accountIdentity` produce the same single key. Proof it fails without the fix — restoring the `login`-first expression:

```
Test case '...testParsedAuthFileAndFetcherAgreeWhenLoginAndUsernameDiffer()' failed (1.538 seconds)
```

One consequence worth flagging: for a file with both fields, this changes `menuBarAccountKey` from `login` to `username`, so a persisted menu bar selection made by a released build would stop resolving. `DirectAuthFile` therefore carries `legacyIdentityKeys` with the previously winning `login`, and `migrateMenuBarSelections` includes it in the Copilot alias list, so those selections migrate onto the canonical key like the filename-based ones.

### Notes

- `@MainActor` reentrancy: the new `await` in the `.copilot` case sits in the same position as the `.codex` one. `previous` is still read once before any suspension and the post-await write is a whole-dictionary replacement for that provider, so no state is read across a suspension point that was not already.
- `AuthFile.quotaLookupKey` and `DirectAuthFile.menuBarAccountKey` keep their exact existing fall-through (`canonicalAccountKey` returns `nil` when there is neither an identity field nor a `github-copilot-*` filename), so non-Copilot and unprefixed-filename behaviour is unchanged. The only intended behaviour change is the field precedence.

`xcodebuild build` succeeds and `xcodebuild test -destination 'platform=macOS'` passes 79/79, no failures (76 on the branch before this commit, plus the 3 new tests).

nguyenphutrong (approved):
The prior blockers are resolved. Provider-scoped Monitor refresh now reconciles the coordinator result, and DirectAuthFileService and CopilotQuotaFetcher share the same username/login/filename identity precedence. Regression coverage exercises differing fields and scoped stale-key reconciliation; the synthetic merge, focused tests, and full suite pass.

PR #470, opened 2026-08-12
feat(logs): display provider in request log view

### LINKED ISSUE #238: Display provider in request log view
### Problem Statement

I go to the log view and cannot see which requests were made to which provider

### Proposed Solution

Display provider in a new column in the request log view

### Alternative Solutions

_No response_

### Feature Category

New AI Provider Integration

### Use Case

_No response_

### Mockups / Examples

_No response_

### Additional Context

_No response_

### Checklist

giaBaoJS:
## Summary

Closes #238

The Requests tab of the Logs screen did not reliably show which provider each request went to: the provider label was tucked inside the model column and was blank whenever detection missed (Qwen, GLM, Grok, DeepSeek, and any request whose path/model was not recognized), and fallback rows only showed the resolved provider inline.

### Before
- Provider text stacked inside the model column, often empty because `RequestLog.provider` was frequently `nil`.
- Provider filter and search only matched the raw stored `provider` field, so entries with a missing provider could never be filtered.

### After
- A dedicated, color-coded **provider badge column** on every request row (shows a localized "Unknown" placeholder when nothing can be derived).
- Provider is now derived in the model layer via `RequestLog.effectiveProvider`:
  `resolvedProvider` (fallback routing) → detected `provider` → inference from the model name (`RequestLog.inferProvider(fromModel:)`). This also fills the provider for previously persisted history entries.
- `ProxyBridge` reuses the same shared inference, which extends detection to Qwen/GLM/Grok/DeepSeek models and fixes resolved Kiro models (`kiro-claude-*`) being misclassified as Claude.
- The provider filter picker, search, and per-provider stats now all use the derived provider, so the column, filter, and stats stay consistent.
- New localization key `logs.provider.unknown` added for en, fr, vi, zh-Hans.

## Changes

- `Quotio/Models/RequestLog.swift` — `effectiveProvider` + `inferProvider(fromModel:)` derivation; per-provider stats aggregate on the derived provider.
- `Quotio/Services/Proxy/ProxyBridge.swift` — replace inline model-name inference with the shared helper.
- `Quotio/Views/Screens/LogsScreen.swift` — provider badge column in `RequestRow`; filter/search use the derived provider.
- `Quotio/Localizable.xcstrings` — `logs.provider.unknown` (en/fr/vi/zh-Hans).
- `QuotioTests/RequestLogProviderTests.swift` — 12 unit tests for the derivation (precedence, model-family inference, kiro-before-claude ordering, unknown fallback, stats aggregation).

## Testing

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED** (all existing suites plus 12 new `RequestLogProviderTests` pass)

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -9659,6 +9659,34 @@
         }
       }
     },
+    "logs.provider.tooltip" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Served by %1$@ · request protocol: %2$@"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Servi par %1$@ · protocole de requête : %2$@"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Được phục vụ bởi %1$@ · giao thức yêu cầu: %2$@"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "由 %1$@ 提供服务 · 请求协议：%2$@"
+          }
+        }
+      }
+    },
     "logs.provider.unknown" : {
       "localizations" : {
         "en" : {
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -9659,6 +9659,34 @@
         }
       }
     },
+    "logs.provider.unknown" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Unknown"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Inconnu"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Không rõ"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "未知"
+          }
+        }
+      }
+    },
     "logs.requestsWillAppear" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/Models/RequestLog.swift
+++ b/Quotio/Models/RequestLog.swift
@@ -163,26 +163,198 @@ nonisolated struct RequestLog: Identifiable, Codable, Hashable, Sendable {
     }
 }
 
+// MARK: - Request Protocol
+
+/// The wire protocol (API shape) a request was made in, derived from its endpoint path.
+///
+/// This is deliberately *not* the provider. `/v1/chat/completions` is the OpenAI-compatible
+/// protocol, and Qwen, GLM, DeepSeek, Grok and every `openai-compatibility` custom provider
+/// speak it too, so the path alone says nothing about which account served the request.
+nonisolated enum RequestProtocol: String, Codable, Hashable, Sendable {
+    case anthropic
+    case openai
+    case gemini
+
+    var displayName: String {
+        switch self {
+        case .anthropic: return "Anthropic"
+        case .openai: return "OpenAI"
+        case .gemini: return "Gemini"
+        }
+    }
+
+    /// The provider label recorded when the protocol is the only available signal.
+    /// These match the values historic logs already contain, so persisted rows keep rendering.
+    var providerLabel: String {
+        switch self {
+        case .anthropic: return "claude"
+        case .openai: return "openai"
+        case .gemini: return "gemini"
+        }
+    }
+
+    /// Detect the protocol from a request path.
+    static func detect(fromPath path: String) -> RequestProtocol? {
+        let lower = path.lowercased()
+
+        // Gemini first: its markers are the most specific, and a Gemini path embeds the model
+        // id (`/v1beta/models/claude-…:generateContent`), which would otherwise trip the
+        // Anthropic check below.
+        if lower.contains("/gemini/") || lower.contains("/google/")
+            || lower.contains("/v1beta/") || lower.contains(":generatecontent") {
+            return .gemini
+        }
+        if lower.contains("/anthropic/") || lower.contains("/claude") || lower.contains("/messages") {
+            return .anthropic
+        }
+        if lower.contains("/openai/") || lower.contains("/chat/completions")
+            || lower.contains("/responses") {
+            return .openai
+        }
+        return nil
+    }
+}
+
 // MARK: - Provider Derivation
 
 nonisolated extension RequestLog {
-    /// The provider that ultimately served the request.
-    /// Prefers the fallback-resolved provider, then the provider detected at request time,
-    /// then infers one from the model name so historical entries also display a provider.
-    var effectiveProvider: String? {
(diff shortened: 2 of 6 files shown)


nguyenphutrong (changes requested):
Thanks for adding the dedicated column and for covering the model-family helper. The synthetic merge into the current `master` builds successfully and all 83 tests pass. There is one blocking data-flow issue, though: the new inference does not run for the common OpenAI-compatible request path, so the column still reports the API protocol rather than the provider/model family this PR says it identifies.

For example, a `POST /v1/chat/completions` request with model `qwen3-coder-plus` is assigned `provider = "openai"` by `ProxyBridge.extractMetadata` before the body is inspected. Because `inferProvider` only runs when that value is nil, the stored log remains `openai`; `effectiveProvider` then prefers that stored value over model inference. The badge, filter, search, and stats therefore all show **OpenAI**, not **Qwen**. The same applies to GLM, Grok, DeepSeek, Kiro-on-an-OpenAI-compatible endpoint, and already-persisted rows with a protocol-derived provider. This contradicts the PR's stated improvement for those families and does not reliably answer issue #238's “which requests were made to which provider.”

Please first define the intended meaning of this field (serving/credential provider versus request protocol/model family), then apply one consistent precedence across extraction and `effectiveProvider`. Be careful that model names alone cannot identify a hosting aggregator (for example, Antigravity or Copilot can serve Claude-prefixed models), while fallback routes already have authoritative `resolvedProvider`. Add integration-style coverage for endpoint + model combinations such as `/v1/chat/completions` + Qwen/GLM and for an aggregator-hosted model, rather than only testing the standalone inference helper.

Once the displayed provider is authoritative and consistent, the new column itself looks suitable.

giaBaoJS:
Thanks — you were right, and the trace confirms it. Pushed 8818c26.

## Data-flow trace of `ProxyBridge.extractMetadata`

Old order (ProxyBridge.swift:613-643 before this commit): path detection first, model only consulted `if provider == nil`. Per request shape the proxy actually serves (agents are configured with `proxyURL = endpoint + "/v1"`, `ANTHROPIC_BASE_URL = endpoint` — AgentSetupViewModel.swift:106, AgentConfigurationService.swift:873):

| Request | path match | stored `provider` | model consulted? |
|---|---|---|---|
| `POST /v1/chat/completions` + `qwen3-coder-plus` | `/chat/completions` | **`openai`** | **no** |
| `POST /v1/chat/completions` + `gpt-5.2` | `/chat/completions` | `openai` | no |
| `POST /v1/messages` + `claude-sonnet-4-5` | none | `claude` | yes (inferred) |
| `POST /v1/responses` + `gpt-5.1-codex` | none | `openai` | yes (inferred) |
| `POST /v1beta/models/gemini-2.5-pro:generateContent` | none (`/gemini/` needs a trailing slash) | `nil` — body has no `model` either | n/a |
| `POST /copilot/v1/chat/completions` | `/chat/completions` wins over `/copilot/` | **`openai`** | no |

So every OpenAI-compatible request — the most common one — was stamped `openai` from the protocol, and `effectiveProvider` (`resolvedProvider ?? provider ?? infer(model)`) preferred that stored value, so Qwen/GLM/Grok/DeepSeek/Kiro-on-an-OpenAI-endpoint all rendered as OpenAI in badge, filter, search and stats. Copilot was worse than you described: the `/copilot/` branch was unreachable for its normal endpoint.

## Field definition and precedence

The field means **the provider that served the request** (issue #238: "which requests were made to which provider"). The protocol is now a separate concept rather than an overloaded value in the same field: `RequestProtocol` (`anthropic` / `openai` / `gemini`, RequestLog.swift) is derived from the endpoint and exposed as `RequestLog.requestProtocol`; it is shown in the badge tooltip, never as the provider.

One precedence, applied identically at capture time (`RequestLog.deriveProvider(path:model:)`, called by `extractMetadata`) and at read time (`effectiveProvider`):

1. **`resolvedProvider`** — the fallback chain records the `AIProvider` it actually dialled; authoritative, unchanged.
2. **A provider-scoped path segment** (`/copilot/`, `/kiro`, `codewhisperer`) — your caveat: aggregators serve other vendors' model families, so a model name must not override a host we can name outright.
3. **The model family** — the only signal that separates Qwen/GLM/DeepSeek/Grok from plain OpenAI on a shared `/v1/chat/completions`.
4. **The protocol label** (`openai`/`claude`/`gemini`) as a last resort, so an unclassifiable model reads exactly as it did before rather than becoming "Unknown".

Why the model outranks the protocol but not a named host: a protocol label is evidence about the *wire format only* and is shared by every `openai-compatibility` provider, whereas `/copilot/` names a host directly. Where model names are ambiguous — an aggregator serving someone else's family — the host wins at step 2 or via an aggregator prefix inside step 3.

**Already-persisted rows**: `effectiveProvider` treats a stored `claude`/`openai`/`gemini` as protocol-only and re-ranks it through steps 3-4 on read, so a legacy row `{provider: "openai", model: "qwen3-coder-plus"}` now renders Qwen without a migration, while `{provider: "openai", model: "gpt-5.2"}` stays OpenAI. Labels that name a real provider (`kiro`, `copilot`, `antigravity`, …) are returned as-is. `canonicalProviderID` also folds the two vocabularies the field has carried — path labels like `copilot` and `AIProvider` raw values like `github-copilot` (written by `resolvedProvider`) — onto one id so badge, filter and stats group them together.

## Model families

Taken from ids this app actually emits, not invented prefixes:
- aggregator prefixes: `cline-pass/…` (CustomProviderSheet.swift:51-70), `kiro-…` / `codewhisperer` and `gemini-claude-…` = Antigravity-hosted Claude (AgentModels.swift:229-233, FallbackModels.swift:107-108, and the same rule FallbackSheets.swift:159-181 already uses for icons);
- generic `AIProvider.rawValue` prefix match, mirroring `FallbackSheets.providerFromModel`;
- families `claude/opus/sonnet/haiku`, `gemini`, `gpt|o1|o3|o4|codex`, and `qwen`, `glm`, `grok`, `deepseek`, `kimi`, `minimax`, `mimo` (the last four are the ClinePass catalogue and `buildOpenCodeModelConfig` at AgentConfigurationService.swift:1298-1305).

Unknown families still return `nil` — a guess is worse than falling through to the protocol label.

## Coverage

`QuotioTests/RequestLogProviderTests.swift` now drives endpoint + model combinations rather than only the helper: `/v1/chat/completions` with Qwen/GLM/Grok/DeepSeek/Kimi, with `gpt-5.2` and `o3-mini` (genuine OpenAI still reads OpenAI), `/v1/responses` + `gpt-5.1-codex`, `/v1/messages` + Claude, `/v1beta/…:generateContent`, aggregator-hosted `/copilot/v1/chat/completions` + `claude-3.5-sonnet` and `gemini-claude-opus-4-6-thinking` and `cline-pass/qwen3.7-max`, plus persisted-row tests including a JSON encode/decode round trip. `testQwenOverOpenAIEndpointReachesBadgeFilterSearchAndStats` asserts your exact scenario across badge text, the filter predicate, the search predicate and `RequestStats.byProvider`.

New string `logs.provider.tooltip` localized for en/fr/vi/zh-Hans.

`xcodebuild build` and `xcodebuild test -destination 'platform=macOS'` both pass — 91 tests, 0 failures.

nguyenphutrong (approved):
The provider/protocol blocker is resolved. Shared OpenAI-compatible paths now inspect the model, resolvedProvider remains authoritative, named aggregators outrank model-family inference, and persisted protocol-only rows are re-ranked consistently. Badge, filter, search, and stats share the same effective provider. The synthetic merge builds and the full suite passes.

PR #471, opened 2026-08-12
feat(auth): preserve unknown JSON fields in auth file structs

### LINKED ISSUE #239: feat: Future-proof auth file structs to preserve unknown JSON fields
## Background

PR #210 fixed a bug where user-defined fields (`prefix`, `project_id`, `proxy_url`) in auth JSON files were being deleted during token refresh. The fix added these specific fields to the `Codable` structs.

However, if CLIProxyAPI adds more user-defined fields in the future, the same issue will recur until those fields are manually added to Quotio's structs.

## Problem

Current approach requires:
1. CLIProxyAPI adds a new field (e.g., `custom_header`)
2. Users report data loss
3. Quotio must release an update adding the field to structs
4. Users must update Quotio

This creates a maintenance burden and potential data loss window.

## Proposed Solution

Implement a "catch-all" mechanism to preserve unknown JSON fields during decode→modify→encode cycles.

### Option A: AnyCodable Dictionary
```swift
nonisolated struct AntigravityAuthFile: Codable, Sendable {
    var accessToken: String
    let email: String
    // ... existing fields ...
    
    // Preserve any unknown fields from CLIProxyAPI
    var additionalFields: [String: AnyCodable]?
    
    init(from decoder: Decoder) throws {
        // Decode known fields normally
        // Capture remaining keys into additionalFields
    }
    
    func encode(to encoder: Encoder) throws {
        // Encode known fields
        // Merge additionalFields back into output
    }
}
```

### Option B: Raw JSON Preservation
```swift
// Store original JSON, only modify specific fields
private var _rawJSON: [String: Any]

func encode(to encoder: Encoder) throws {
    var updated = _rawJSON
    updated["access_token"] = accessToken
    // ... update only modified fields
    // Encode updated dictionary
}
```

## Affected Files

- `Quotio/Services/Antigravity/AntigravityQuotaFetcher.swift` → `AntigravityAuthFile`
- `Quotio/Services/QuotaFetchers/OpenAIQuotaFetcher.swift` → `CodexAuthFile`

## Complexity

**Medium** - Requires custom `Codable` implementation and potentially an `AnyCodable` type (or dependency).

## Trade-offs

| Approach | Pros | Cons |
|----------|------|------|
| Current (explicit fields) | Simple, type-safe | Manual updates needed |
| AnyCodable | Automatic preservation | More complex, needs AnyCodable impl |
| Raw JSON | Full fidelity | Loses type safety for known fields |

## Acceptance Criteria

- [ ] Unknown JSON fields in auth files survive token refresh
- [ ] Known fields remain type-safe and accessible
- [ ] No performance regression on auth file operations
- [ ] Unit tests covering field preservation

---

*Originated from PR #210 review*

giaBaoJS:
## Motivation

Issue #239: auth file structs only round-trip the JSON fields they explicitly declare. PR #210 fixed the loss of `prefix`, `project_id`, and `proxy_url` by adding those specific fields, but any field CLIProxyAPI (or the user) adds in the future would be silently dropped again whenever Quotio decodes and re-encodes an auth file — a recurring data-loss window until each new field is manually added.

While auditing the current write paths, the disk rewrites under `~/.cli-proxy-api` were already converted to `JSONSerialization` read-modify-write (#312), but the structs themselves still lose unknown fields when encoded — and one live path still does exactly that today: `CodexCLIQuotaFetcher` decodes the native Codex credential (`CodexCLIAuthFile`) from the Keychain, mutates tokens, and writes back `JSONEncoder().encode(auth)`, dropping any field the struct doesn't model.

## Approach

Struct-level catch-all (Option A from the issue, with a typed `JSONValue` instead of an untyped `AnyCodable`):

- **`Quotio/Models/JSONValue.swift`** (new): a `nonisolated enum JSONValue: Codable, Equatable, Sendable` representing arbitrary JSON (string/int/double/bool/object/array/null), plus helpers `JSONValue.unknownFields(from:excluding:)` and `JSONValue.encodeUnknownFields(_:to:)` built on a dynamic `JSONCodingKey`.
- **`AntigravityAuthFile`**, **`CodexAuthFile`**, **`CodexCLIAuthFile`**, and **`CodexCLITokens`** get an `additionalFields: [String: JSONValue]` catch-all with custom `init(from:)` / `encode(to:)`: typed fields decode/encode exactly as before, and every unrecognized key is captured on decode and re-emitted on encode.

Known fields stay fully type-safe and the JSON produced for files without unknown fields is unchanged, so this is behavior-preserving for existing data. The Codex Keychain refresh path now preserves unknown fields automatically with no call-site changes, and any future code that round-trips these structs is safe by construction.

## Test evidence

New `QuotioTests/AuthFileUnknownFieldsTests.swift` (5 tests) round-trips JSON containing unknown fields of every shape — `custom_header` string, int, double, bool, `null`, arrays, and nested objects — through decode → mutate typed fields → encode for all four structs, including unknown keys nested inside `tokens` (the object rewritten by the Keychain refresh path), and asserts an unmodified round trip is byte-for-byte lossless at the JSON level:

```
Test case 'AuthFileUnknownFieldsTests.testAntigravityAuthFileDecodesKnownFieldsWithoutUnknownFields()' passed
Test case 'AuthFileUnknownFieldsTests.testAntigravityAuthFilePreservesUnknownFieldsAcrossRoundTrip()' passed
Test case 'AuthFileUnknownFieldsTests.testCodexAuthFilePreservesUnknownFieldsAcrossRoundTrip()' passed
Test case 'AuthFileUnknownFieldsTests.testCodexCLIAuthFilePreservesUnknownFieldsIncludingNestedTokens()' passed
Test case 'AuthFileUnknownFieldsTests.testRoundTripWithoutModificationIsLossless()' passed
```

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` → **BUILD SUCCEEDED**
- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS'` → **TEST SUCCEEDED** (full suite)

Closes #239

### DIFF
--- a/Quotio/Models/JSONValue.swift
+++ b/Quotio/Models/JSONValue.swift
@@ -15,12 +15,29 @@ import Foundation
 /// decode → modify → encode cycle instead of being silently dropped.
 nonisolated indirect enum JSONValue: Codable, Equatable, Sendable {
     case string(String)
-    case int(Int)
+    /// Any JSON number Foundation can materialize exactly.
+    ///
+    /// `Decimal` is used rather than `Int`/`Double` because it is a base-10 type:
+    /// it keeps integers larger than `Int64.max` (e.g. `9223372036854775809`) and
+    /// decimal fractions (e.g. `0.1`) byte-identical across a decode → encode cycle,
+    /// where `Double` would round them to the nearest binary value.
+    case number(Decimal)
+    /// Finite literals outside `Decimal`'s exponent range (roughly |exponent| > 128)
+    /// but still representable as `Double`, e.g. `1e308`.
     case double(Double)
     case bool(Bool)
     case object([String: JSONValue])
     case array([JSONValue])
     case null
+
+    /// Thrown when a JSON number literal falls outside every numeric type Foundation
+    /// can materialize (e.g. `1e400`, which overflows both `Decimal` and `Double`).
+    ///
+    /// RFC 8259 §6 explicitly allows a parser to limit the range it accepts, and
+    /// Foundation does: `JSONDecoder` reports "not representable in Swift" and
+    /// `JSONSerialization` rejects the whole document. Unknown fields carrying such a
+    /// literal are therefore skipped instead of aborting the auth-file decode.
+    nonisolated struct UnrepresentableNumberError: Error, Equatable {}
 
     init(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
@@ -28,9 +45,9 @@ nonisolated indirect enum JSONValue: Codable, Equatable, Sendable {
             self = .null
         } else if let value = try? container.decode(Bool.self) {
             self = .bool(value)
-        } else if let value = try? container.decode(Int.self) {
-            self = .int(value)
-        } else if let value = try? container.decode(Double.self) {
+        } else if let value = try? container.decode(Decimal.self) {
+            self = .number(value)
+        } else if let value = try? container.decode(Double.self), value.isFinite {
             self = .double(value)
         } else if let value = try? container.decode(String.self) {
             self = .string(value)
@@ -39,10 +56,7 @@ nonisolated indirect enum JSONValue: Codable, Equatable, Sendable {
         } else if let value = try? container.decode([JSONValue].self) {
             self = .array(value)
         } else {
-            throw DecodingError.dataCorruptedError(
-                in: container,
-                debugDescription: "Unsupported JSON value"
-            )
+            throw UnrepresentableNumberError()
         }
     }
 
@@ -50,7 +64,7 @@ nonisolated indirect enum JSONValue: Codable, Equatable, Sendable {
         var container = encoder.singleValueContainer()
         switch self {
         case .string(let value): try container.encode(value)
-        case .int(let value): try container.encode(value)
+        case .number(let value): try container.encode(value)
         case .double(let value): try container.encode(value)
         case .bool(let value): try container.encode(value)
         case .object(let value): try container.encode(value)
@@ -79,11 +93,19 @@ nonisolated extension JSONValue {
     ///
     /// Call from a custom `init(from:)` after decoding the typed fields to
     /// capture unrecognized fields so they can be re-emitted on encode.
+    ///
+    /// A field carrying a number literal Foundation cannot represent is skipped
+    /// rather than propagated, so one exotic unknown value can never make an auth
+    /// file undecodable and break token refresh.
     static func unknownFields(from decoder: Decoder, excluding knownKeys: Set<String>) throws -> [String: JSONValue] {
         let container = try decoder.container(keyedBy: JSONCodingKey.self)
         var fields: [String: JSONValue] = [:]
         for key in container.allKeys where !knownKeys.contains(key.stringValue) {
-            fields[key.stringValue] = try container.decode(JSONValue.self, forKey: key)
+            do {
+                fields[key.stringValue] = try container.decode(JSONValue.self, forKey: key)
+            } catch is UnrepresentableNumberError {
+                continue
+            }
         }
         return fields
     }
--- a/Quotio/Models/JSONValue.swift
+++ b/Quotio/Models/JSONValue.swift
@@ -1,0 +1,99 @@
+//
+//  JSONValue.swift
+//  Quotio - CLIProxyAPI GUI Wrapper
+//
+//  Type-safe representation of arbitrary JSON values, used to preserve
+//  unknown fields when auth files are decoded, modified, and re-encoded.
+//
+
+import Foundation
+
+/// A type-safe representation of any JSON value.
+///
+/// Auth file structs use this as a catch-all so fields Quotio does not model
+/// (e.g. fields added by newer CLIProxyAPI versions) survive a
+/// decode → modify → encode cycle instead of being silently dropped.
+nonisolated indirect enum JSONValue: Codable, Equatable, Sendable {
+    case string(String)
+    case int(Int)
+    case double(Double)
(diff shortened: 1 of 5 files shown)


nguyenphutrong (changes requested):
Thanks for addressing unknown-field preservation across the auth DTOs. The synthetic merge into the current `master` builds successfully, and all 76 tests pass.

One correctness blocker remains: `JSONValue` decodes every unknown number as `Int` and then `Double`. This does not preserve arbitrary valid JSON numbers. For example, an unknown value of `9223372036854775809` is re-encoded as `9.223372036854776e+18`, changing its exact value; a valid exponent such as `1e400` makes the containing auth DTO fail to decode entirely because it fits neither `Int` nor finite `Double`.

This affects the native Codex Keychain token-refresh path, which decodes and re-encodes the complete `CodexCLIAuthFile`, so valid unknown fields can still be corrupted or prevent refresh. That conflicts with #239's requirement that unknown JSON fields survive token refresh.

Please preserve unknown numeric values without narrowing them to `Int`/`Double`, and add regression tests covering an integer beyond `Int64` and a large exponent. Once those cases round-trip without changing the numeric value or failing the auth-file decode, this should be ready for another review.

giaBaoJS:
Fixed in 3f29720. Unknown numbers are no longer narrowed to `Int`/`Double`.

**Approach:** `JSONValue` now holds numbers as `case number(Decimal)`, with `case double(Double)` as a fallback for finite literals outside `Decimal`'s exponent range. `Decimal` is a base-10 type, so a JSON literal round-trips byte-identically instead of being routed through a binary `Double`.

I tested each candidate on both values before picking (Swift 6.2.3 / macOS 26.5, `JSONDecoder` + `JSONEncoder`):

| literal | old `Int`→`Double` | `JSONSerialization` | `Decimal` |
|---|---|---|---|
| `9223372036854775809` | `9.223372036854776e+18` | exact | **exact** |
| `1e400` | decode throws | rejects the **whole document** ("Number wound up as NaN") | decode throws |
| `1e-400` | silently `0` | silently `0` | decode throws |
| `0.1` | `0.1` (binary-rounded) | `0.1` | **exact** |

So `JSONSerialization` was ruled out — it is strictly worse (it fails the entire auth file on `1e400`, and silently zeroes `1e-400`). `Decimal` fixes the >`Int64` case exactly, and also stops `0.1`-style fractions being re-emitted through a binary `Double`.

**On `1e400` specifically:** no Foundation numeric type can hold it — `Decimal`, `Double`, `Int` and `UInt64` all fail with `numberIsNotRepresentableInSwift(parsed: "1e400")`, and there is no public API to read a number's raw lexical form from a `Decoder`. Preserving it verbatim would mean replacing Foundation's JSON stack with a hand-rolled parser *and* serializer across every auth call site — and the emitted file would then be unreadable by the `JSONSerialization`-based auth paths already in the app (e.g. `ClaudeCodeQuotaFetcher.swift:201-211`). RFC 8259 §6 explicitly permits a parser to limit accepted number range, which is what Foundation does here.

What I fixed instead is the harm you identified: **the decode must never fail.** `JSONValue.unknownFields(from:excluding:)` now catches a dedicated `JSONValue.UnrepresentableNumberError` and skips only that one unknown key, so the auth file still decodes and token refresh still completes. That single field is dropped — the same outcome as `master`, which drops all unknown fields — rather than breaking refresh entirely. Worth noting the `1e400` decode failure was introduced by this PR's catch-all; on `master` the document parses fine because the unknown value is never materialized (verified by decoding a DTO that ignores the key).

**Regression tests** (`QuotioTests/AuthFileUnknownFieldsTests.swift`) run the real `CodexCLIAuthFile` decode → mutate `tokens` → encode cycle from `persistNativeKeychainRefresh`, not just a `JSONValue` unit test:

- `testNativeKeychainRefreshPreservesIntegerBeyondInt64` — `9223372036854775809` in both a top-level unknown field and an unknown field inside `tokens`; asserts the encoded bytes contain `"install_id":9223372036854775809` and `"token_serial":9223372036854775809`, and explicitly that `9.223372036854776e+18` is absent. Also covers `0.1`.
- `testNativeKeychainRefreshSurvivesUnrepresentableExponent` — `1e400` and `1e-400` at top level and inside `tokens`; the refresh completes, rotated tokens and every sibling unknown field (including a `9223372036854775809`) survive, and only the two overflowing keys are absent.
- Two `JSONValue` tests cover the `Decimal`/`Double`/`String`/`Bool` classification (numbers must not swallow `"123"` or `true`) and the distinct error type.

`xcodebuild build` and `xcodebuild test -destination 'platform=macOS'` both pass (`** TEST SUCCEEDED **`, 75 cases), including the pre-existing lossless round-trip test.

nguyenphutrong (changes requested):
Decimal fixes the beyond-Int64 precision case and an unrepresentable exponent no longer aborts the whole DTO decode, but one data-integrity blocker remains. JSONValue.unknownFields catches UnrepresentableNumberError and skips the key, so valid unknown fields such as 1e400/1e-400 are deleted during decode → token mutation → encode; the new regression test explicitly expects that deletion. Issue #239 requires unknown fields to survive token refresh. Please preserve the raw numeric token/value through this path, or explicitly narrow the feature contract rather than presenting it as full unknown-field preservation. The synthetic merge and targeted tests pass, but they currently encode the incorrect data-loss behavior.

nguyenphutrong:
Thank you for the contribution. This PR is being closed because the proposed catch-all Codable layer is no longer aligned with Quotio's current auth-file handling plan. The linked issue will remain open so it can be revisited with an implementation that fits the current architecture.

Issue #352, opened 2026-03-08
Cannot delete account


zwmjohn-cool:
### Problem Statement

Cannot delete account

### Proposed Solution

add delete account botton

### Feature Category

New AI Provider Integration

### How important is this to you?

🔴 Critical - Blocking my workflow

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [ ] I am willing to contribute to implementing this feature

Issue #29, opened 2025-12-29
Quotio requests access to all filepath locations


vre:
Quotio requests access to all possible file paths on Mac. I don't see a reason why Quotio would need to traverse around the filesystem. Or even access any of the files in the systems unless I specifically request it to open some key file etc.

<img width="272" height="277" alt="Image" src="https://github.com/user-attachments/assets/38fff49f-20a5-465b-b3b6-1b813e8150b7" />
<img width="272" height="241" alt="Image" src="https://github.com/user-attachments/assets/d424917d-586d-485d-bd90-8729d2618f9a" />
<img width="269" height="216" alt="Image" src="https://github.com/user-attachments/assets/66799018-56fc-40d3-8962-741bc7c0003c" />
<img width="264" height="213" alt="Image" src="https://github.com/user-attachments/assets/f00c2be0-df56-4418-8c76-d1a21822a3c1" />

nguyenphutrong:
Hi @vre, thank you for reporting this security concern! 😊

I understand your concern - Quotio shouldn't need access to all file paths on your system. 

**Root Cause Analysis:**
I've investigated the codebase and found several problematic access patterns:
1. **Cross-app Library access**: Reading Cursor/Trae databases in `~/Library/Application Support/`
2. **Broad binary scanning**: Scanning multiple system paths (`/usr/bin`, `/usr/local/bin`, etc.)
3. **Home directory enumeration**: Checking various hidden directories (`~/.cargo/bin`, `~/.npm-global/bin`, etc.)

**Fix Plan:**
- ✅ Remove cross-application Library access
- ✅ Limit CLI binary detection to user-specified paths only  
- ✅ Implement proper sandbox configuration
- ✅ Add user consent for file access

I'll be implementing these fixes in the coming days. The goal is to only request access to:
- App-specific directories (`~/Library/Application Support/Quotio/`)
- User-selected configuration files
- Explicitly approved CLI binary paths

Thanks again for bringing this to my attention! Security and privacy are top priorities for Quotio. 🛡️

PR #472, opened 2026-08-12
fix(providers): allow deleting accounts imported from Cursor and Trae

### LINKED ISSUE #213: Unable to delete accounts imported from Cursor
### What happened?

Unable to delete accounts imported from Cursor

### Steps to reproduce

1. open Quotio 
2. click Providers
3. select Cursor

### Expected behavior

Import Curosr account 

### Severity

### Affected component

Dashboard

### Provider (if applicable)

None

### CLI Agent (if applicable)

None

### macOS Version

26.2 (25C56)

### Quotio Version

0.7.6

### App Mode

Full Mode (Proxy running)

### Screenshots

_No response_

### Logs (optional)

```shell

```

### Additional context

_No response_

giaBaoJS:
## Problem

Accounts imported from Cursor (and Trae) via **Providers → Scan for IDEs** cannot be deleted (#213, likely also the report in #352):

- In Local Proxy / quota-only mode these accounts render as auto-detected rows with a hardcoded `canDelete: false` (`AccountRowData.from(provider:accountKey:)`), so `AccountRow` shows no trash button and no context-menu delete entry.
- In Monitor mode they surface as quota-derived accounts (`MonitorRefreshCoordinator.makeQuotaDerivedAccount`) which default to `canDelete: false`, and `deleteMonitorAccount` guards on that flag, so deletion is refused there too.
- The imported quota data is persisted in two places — UserDefaults (`persisted.ideQuotas`) and the Monitor snapshot (`Monitor/snapshots-v1.json`) — with no cleanup path, so the account survives every restart with no way to remove it.

## Fix

- Allow delete for auto-detected Cursor/Trae rows and for their quota-derived Monitor accounts (gated on `usesBrowserAuth`, so credential-backed placeholders that would just resurrect on the next discovery pass stay non-deletable).
- New `QuotaViewModel.deleteAutoDetectedAccount(provider:accountKey:)` and a hardened `deleteMonitorAccount` remove **all** backing state: the `providerQuotas` entry (dropping the provider key when it empties), the persisted IDE quotas in UserDefaults, the Monitor snapshot, and the menu bar selection — so the account does not resurrect on the next launch or refresh.
- `refreshAutoDetectedProviders()` (invoked by the generic Providers toolbar refresh) no longer re-imports Cursor/Trae from the local IDE databases. Those providers are documented as import-only-via-explicit-scan (issue #29 consent model), and re-importing here would both bypass that consent and resurrect deleted accounts. Scoped per-provider/per-account refresh still works while an account is imported, and "Scan for IDEs" re-imports on demand.

## Tests

Added to `QuotioTests/MonitorRuntimeTests`:

- `testQuotaDerivedIDEAccountsAllowDeletion` — Cursor/Trae quota-derived Monitor accounts are deletable; credential-backed (`claude`) and API-key (`amp`) placeholders remain non-deletable.
- `testAutoDetectedAccountRowsAllowDeleteOnlyForIDEProviders` — Providers-screen rows offer delete for Cursor/Trae only (not Devin/Grok).
- `testSnapshotStoreDropsDeletedIDEAccountAcrossReload` — after the Cursor entry is removed and the snapshot re-persisted, a fresh load does not resurrect the account while sibling providers are preserved.

## Verification

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` — succeeds
- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS'` — all tests pass

Fixes #213

### DIFF
--- a/Quotio/Models/Models.swift
+++ b/Quotio/Models/Models.swift
@@ -217,7 +217,19 @@ nonisolated enum AIProvider: String, CaseIterable, Codable, Identifiable, Sendab
             return false
         }
     }
-    
+
+    /// Whether this provider's accounts are imported from a local IDE database by the
+    /// explicit "Scan for IDEs" flow (issue #29) instead of being authenticated inside
+    /// Quotio. Such an account owns no Quotio credential, so it exists only as imported
+    /// quota data and deleting that data deletes the account (issue #213).
+    ///
+    /// Derived from the existing traits rather than listing cases again, so a new IDE
+    /// edition only has to opt into `usesBrowserAuth` / `supportsManualAuth` to inherit
+    /// the same import and delete behaviour.
+    var isImportedFromLocalIDE: Bool {
+        usesBrowserAuth && !supportsManualAuth
+    }
+
     /// Map provider to CLI agent (if applicable)
     var cliAgent: CLIAgent? {
         switch self {
--- a/Quotio/Services/Monitor/MonitorRuntime.swift
+++ b/Quotio/Services/Monitor/MonitorRuntime.swift
@@ -283,6 +283,41 @@ actor MonitorSnapshotStore {
             Log.quota("Failed to persist Monitor snapshot: \(error.localizedDescription)")
         }
     }
+
+    /// Drop a single account from the persisted snapshot, leaving every other entry as
+    /// it is on disk.
+    ///
+    /// `store(_:)` replaces the whole file with the caller's in-memory quotas, which is
+    /// only correct while Monitor mode is active. Deleting an imported IDE account is
+    /// reachable from the other modes too, where the in-memory quotas belong to that
+    /// mode, so the delete path needs this targeted removal instead (issue #213).
+    ///
+    /// - Returns: `true` when an entry was found and the file was rewritten.
+    @discardableResult
+    func removeAccount(provider: AIProvider, accountKey: String) -> Bool {
+        var quotas = load()
+        guard var accountQuotas = quotas[provider] else { return false }
+
+        // Match the identity Monitor itself dedupes on (`MonitorAccount.deduplicationKey`):
+        // the Monitor row carries a trimmed account key while the snapshot key is stored
+        // verbatim, so an exact-only match could leave the entry behind.
+        let normalized = accountKey.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        let matchingKeys = accountQuotas.keys.filter {
+            $0 == accountKey || $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() == normalized
+        }
+        guard !matchingKeys.isEmpty else { return false }
+
+        for key in matchingKeys {
+            accountQuotas.removeValue(forKey: key)
+        }
+        if accountQuotas.isEmpty {
+            quotas.removeValue(forKey: provider)
+        } else {
+            quotas[provider] = accountQuotas
+        }
+        store(quotas)
+        return true
+    }
 }
 
 actor MonitorAccountDiscovery {
@@ -636,14 +671,15 @@ actor MonitorRefreshCoordinator {
         source: MonitorAccountSource,
         disabledIDs: Set<String>
     ) -> MonitorAccount {
-        // Cursor/Trae accounts exist only as imported quota snapshots (no owned
-        // credential), so deleting the snapshot fully removes them (issue #213).
+        // Accounts imported from a local IDE (Cursor, Trae) exist only as imported quota
+        // data with no owned credential, so removing that data fully deletes them and
+        // they will not reappear on the next discovery pass (issue #213).
         var account = MonitorAccount.make(
             provider: provider,
             accountKey: accountKey,
             displayName: displayName,
             source: source,
-            canDelete: provider.usesBrowserAuth
+            canDelete: provider.isImportedFromLocalIDE
         )
         account.isDisabled = disabledIDs.contains(account.id)
         return account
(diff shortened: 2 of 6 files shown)


nguyenphutrong (changes requested):
Thanks for adding the missing deletion affordance and covering both Cursor and Trae. The synthetic merge into the current `master` builds successfully. In the full test run, 73/74 tests passed; the unrelated concurrency/coalescing test failed once and then passed immediately when rerun in isolation.

One persistence blocker remains. `deleteAutoDetectedAccount` rewrites the Monitor snapshot only inside `if modeManager.isMonitorMode`. In the UI, however, this method handles `.autoDetected` rows outside Monitor mode; Monitor rows are dispatched to `deleteMonitorAccount` first. Consequently, deleting a Cursor/Trae account in Local Proxy mode updates the in-memory quotas and UserDefaults but leaves any existing `Monitor/snapshots-v1.json` entry untouched.

A concrete sequence is:

1. Import Cursor while in Monitor mode and let the Monitor snapshot persist it.
2. Switch to Local Proxy mode and delete the auto-detected Cursor row.
3. Switch back to Monitor mode.

`initializeQuotaOnlyMode()` then bootstraps from the stale Monitor snapshot and restores the deleted account. This conflicts with the PR's stated requirement to remove all backing state and prevent resurrection.

Please remove the account from the Monitor snapshot regardless of the currently active mode (without replacing it with unrelated mode-specific quota data), and add a regression test covering deletion outside Monitor mode followed by a Monitor snapshot/bootstrap reload. The current snapshot-store test only manually edits and stores a dictionary, so it does not exercise this missing deletion path.

Non-blocking: `usesBrowserAuth` currently doubles as “is an explicitly imported IDE provider” and “is deletable.” An IDE-specific capability/set would make the intent less fragile, though this is not required to resolve the blocker.

giaBaoJS:
Thanks — you were right, and the fix is pushed in `471deed`.

## Blocker: the snapshot is now cleared in every mode

`MonitorSnapshotStore.removeAccount(provider:accountKey:)` drops a single entry and rewrites only that file; `MonitorRefreshCoordinator.forgetSnapshotAccount(provider:accountKey:)` wraps it. Both `deleteAutoDetectedAccount` and `deleteMonitorAccount` call it unconditionally, outside any `isMonitorMode` branch.

It is deliberately *not* a `finish(quotas:)` rewrite: outside Monitor mode `providerQuotas` holds that mode's quota data, so rewriting the whole snapshot from it would replace the Monitor snapshot with unrelated data — exactly what you asked me to avoid. `finish(quotas:)` stays behind the `isMonitorMode` check, where it is correct.

## Store inventory for an account imported from Cursor/Trae

| # | Backing store | Cleared by | Mode-independent? |
|---|---|---|---|
| 1 | in-memory `providerQuotas[provider][accountKey]` | `removeQuotaEntry` | yes |
| 2 | UserDefaults `persisted.ideQuotas` | `removeQuotaEntry` → `savePersistedIDEQuotas()` | yes |
| 3 | `Monitor/snapshots-v1.json` | `forgetSnapshotAccount` → `MonitorSnapshotStore.removeAccount` | **yes (this was the bug)** |
| 4 | `Monitor/accounts-v1.json` → `disabledAccountIDs` | `forgetSnapshotAccount` (only writes if the id is actually present) | yes |
| 5 | `Monitor/accounts-v1.json` → `accounts` | n/a — quota-derived rows are never written by `MonitorMetadataStore.saveAccount`; `deleteMonitorAccount` still clears it via `deleteOwnedAccount` for credential-backed accounts | yes |
| 6 | Keychain Monitor credential | n/a — these accounts own no credential; `deleteMonitorAccount` still calls `deleteOwnedAccount` | yes |
| 7 | menu bar selection (`MenuBarSettingsManager.selectedItems`) | `syncMenuBarSelection()` → `pruneMenuBarItems()`, which derives valid items from `providerQuotas` | yes |

Store 4 was leftover state I found while enumerating: a Cursor row disabled in Monitor mode and then deleted elsewhere left its id in `disabledAccountIDs`, so a later re-scan would re-import it already disabled.

Reverse direction: nothing re-adds the account before the mode switch either. `refreshAutoDetectedProviders()` excludes IDE-imported providers, and `refreshQuotasUnified` / `refreshAllQuotas` / `refreshQuotasDirectly` never touch Cursor/Trae — re-import stays gated on an explicit "Scan for IDEs" (#29).

## Regression test — your exact sequence

`testDeletingIDEAccountOutsideMonitorModeSurvivesMonitorBootstrap` runs the three steps against real files in a temp directory, with an injected snapshot store, metadata store and empty credential vault:

1. Monitor mode persists Cursor (plus an unrelated Codex entry) via `coordinator.finish(quotas:)`; a `bootstrap()` asserts the account really is restored, so the test fails loudly if the precondition stops holding.
2. Deletion outside Monitor mode calls the production `forgetSnapshotAccount` — the call `deleteAutoDetectedAccount` makes.
3. A fresh coordinator calls `bootstrap()`, which is what `initializeQuotaOnlyMode()` calls. Asserts `quotas[.cursor] == nil`, no `.cursor` account in `batch.accounts`, and that the Codex entry still reads 42 — i.e. the snapshot was not clobbered with the deleting mode's data.

Four more: `testForgettingIDEAccountLeavesUnrelatedSnapshotEntriesIntact` (sibling account of the same provider survives), `testSnapshotRemovalMatchesMonitorAccountIdentity` (the Monitor row carries a trimmed key while the snapshot stores it verbatim, so removal matches on `deduplicationKey` identity), `testForgettingIDEAccountClearsItsMonitorDisabledFlag` (store 4), and `testImportedFromLocalIDECapabilityStaysDerivedFromProviderTraits`. The old dictionary-editing test is gone.

## `@MainActor` reentrancy

`removeQuotaEntry(provider:accountKey:)` is synchronous by design and is re-applied *after* `await forgetSnapshotAccount(...)` in both delete paths, so the post-suspension state is re-read rather than trusted from before the `await`.

## Non-blocking: `usesBrowserAuth` overload

Addressed. New `AIProvider.isImportedFromLocalIDE`, used at the three delete/refresh sites instead of `usesBrowserAuth`. It is **derived** (`usesBrowserAuth && !supportsManualAuth`) rather than a second hardcoded case list, so it names the intent without creating a set that can drift — and #479's Trae CN inherits it automatically. `testImportedFromLocalIDECapabilityStaysDerivedFromProviderTraits` pins that.

## Composition with the other open PRs

I trial-merged both on top of this branch:

- **#479 (Trae CN)** — merges cleanly, full suite green. `.traeCn` joins `usesBrowserAuth` and is excluded from `supportsManualAuth`, so `isImportedFromLocalIDE` is true for it and it inherits the delete affordance, the snapshot/metadata cleanup and the auto-refresh exclusion with no extra change. UserDefaults cleanup follows `ideProvidersToSave`, which #479 extends. `isImportedFromLocalIDE` is placed away from #479's hunks so `Models.swift` merges without conflict.
- **#486 (global IDE quota refresh)** — merges cleanly, full suite green. No shared hunks. `refreshImportedIDEQuotas()` merges only into account keys already in `providerQuotas`, and after a delete the key is gone from every store above, so the refresh cannot bring it back; its `testMergeImportedIDEQuotasDoesNotResurrectDeletedAccount` asserts that directly.

## Verification

```
xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build          # ** BUILD SUCCEEDED **
xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'  # ** TEST SUCCEEDED **
```

Full suite green on this branch and on both trial merges. `testCoordinatorCoalescesConcurrentRefreshForProvider` passed in every run this time.

nguyenphutrong (approved):
The stale Monitor-snapshot blocker is resolved. Deletion now removes the specific persisted account in every mode, preserves unrelated snapshot data, clears disabled state, and re-applies in-memory removal after the await. The regression test covers deletion outside Monitor mode followed by bootstrap; the synthetic merge and full suite pass.

PR #473, opened 2026-08-12
feat(codex): add configurable model reasoning effort

### LINKED ISSUE #384: feat: Add support for Codex model_effort configuration
### Problem Statement

Currently, Quotio does not support configuring or displaying the model_effort parameter for Codex models.
The official Codex API requires the model_effort field, but Quotio's UI and proxy config do not provide an option to set it (low / medium / high / xhigh).

### Proposed Solution

Add a model_effort dropdown/input field in the Codex model settings
Support values: low, medium, high, xhigh ....
Pass the model_effort parameter correctly in API requests to Codex

### Alternative Solutions

_No response_

### Feature Category

New AI Provider Integration

### Use Case

_No response_

### Mockups / Examples

_No response_

### Additional Context

_No response_

### Checklist

giaBaoJS:
## Motivation

Quotio hardcodes `model_reasoning_effort = "high"` when generating `~/.codex/config.toml`, so there is no way to pick a different effort level from the app. Codex CLI accepts `minimal`, `low`, `medium`, `high`, and `xhigh` (model-dependent), and users want to tune this per their workflow.

Closes #384

## What changed

- **New `CodexReasoningEffort` enum** (`AgentModels.swift`) with the five values Codex CLI accepts, plus a `defaultEffort` of `.high` that preserves current behavior.
- **`AgentConfiguration.codexReasoningEffort`** carries the selection into config generation; `buildManagedCodexTOML` now emits the selected value instead of the hardcoded string.
- **UI**: the Codex agent configuration sheet gains a "Reasoning Effort" section with a menu picker (proxy mode), styled like the existing Model Slots section. In manual mode, changing the effort regenerates the raw config preview.
- **Round-trip**: `readCodexConfig` now parses `model_reasoning_effort` from an existing `config.toml`, so reopening the sheet pre-selects the currently saved level. The managed-key merge logic already treats `model_reasoning_effort` as a managed top-level key, so user-edited config files keep round-tripping correctly with exactly one effort key emitted.
- **Localization**: new strings added for all catalog languages (en, fr, vi, zh-Hans).

## UI

The Codex configuration sheet now shows a "Reasoning Effort" card between Connection Info and the manual preview/test sections, with a dropdown of Minimal / Low / Medium / High / Extra High and a caption explaining the tradeoff. Default selection is High.

## Testing

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` — succeeds
- `xcodebuild test -scheme Quotio -destination 'platform=macOS'` — all tests pass, including 6 new tests in `QuotioTests/CodexReasoningEffortTests.swift`:
  - managed TOML emits each selected effort value
  - default stays `high` (both enum default and generated TOML)
  - raw values match Codex CLI's accepted set
  - merge with a user-edited config replaces the managed effort key, preserves user keys/sections, and emits the key exactly once

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -1972,6 +1972,34 @@
         }
       }
     },
+    "agents.reasoningEffort.custom" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@ (custom)"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@ (personnalisé)"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@ (tùy chỉnh)"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@（自定义）"
+          }
+        }
+      }
+    },
     "agents.reasoningEffort.high" : {
       "localizations" : {
         "en" : {
@@ -2056,6 +2084,34 @@
         }
       }
     },
+    "agents.reasoningEffort.max" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Max"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Maximum"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Tối đa"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "最大"
+          }
+        }
+      }
+    },
     "agents.reasoningEffort.medium" : {
       "localizations" : {
         "en" : {
@@ -2108,6 +2164,62 @@
           "stringUnit" : {
             "state" : "translated",
             "value" : "最小"
+          }
+        }
+      }
+    },
+    "agents.reasoningEffort.none" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "None"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Aucun"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Không"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "无"
+          }
+        }
+      }
+    },
+    "agents.reasoningEffort.ultra" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Ultra"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Ultra"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Siêu cao"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "超强"
           }
         }
       }
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -1944,6 +1944,202 @@
         }
       }
     },
+    "agents.reasoningEffort" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Reasoning Effort"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Effort de raisonnement"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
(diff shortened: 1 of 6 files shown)


nguyenphutrong (changes requested):
Thank you for adding configurable Codex reasoning effort. I tested this PR merged into the current `master`: the Debug build succeeds and all 77 tests pass. The configuration key and its mapping to Codex are correct, but there are two data-preservation issues that need to be addressed before merging:

1. `readCodexConfig` searches all trimmed lines for `model_reasoning_effort` without tracking TOML sections. A value under a later profile, for example `[profiles.work]`, can therefore be reported as the top-level value. Saving from the sheet may then overwrite the actual top-level setting with the profile value.
2. Codex accepts more values than the five enum cases in this PR. Valid existing values such as `none`, `max`, `ultra`, and non-empty custom values currently decode as `nil`, fall back to `high`, and can be overwritten on save. That makes opening and applying the configuration destructive for valid user-edited configs.

Please make the parser section-aware so it reads only the top-level setting, and preserve/support valid existing Codex values rather than silently replacing them. Regression tests covering a top-level value plus one or more profile values, as well as `none`, `max`, `ultra`, and a custom value, would help lock in the behavior.

giaBaoJS:
Thanks for the careful review — both points were real and both are fixed in 9d2cfdd.

**1. Section-aware parsing**

`readCodexConfig` no longer scans for the key inline. It delegates to a new `parseTopLevelCodexReasoningEffort(from:)` that returns as soon as it hits the first table header, so `[profiles.work]`, `[[mcp_servers]]` and anything else after it can no longer be reported as the top-level value. It follows the same conventions the merge path already uses — `parseTOMLSectionName` for `[table]` / `[[array-of-tables]]`, and exact key comparison like `isCodexManagedTopLevelKey` rather than the old `hasPrefix`, so `model_reasoning_effort_override` and `plan_mode_reasoning_effort` are no longer matched. It additionally skips `#` comments, accepts quoted keys and literal (`'…'`) strings, strips trailing comments from the value, and tracks `"""` / `'''` state so a key that appears inside a multi-line string body is not read.

**2. Accepted values — evidence**

The set is genuinely open, so I treated it as open rather than extending a fixed list by guesswork. From `codex-rs/protocol/src/openai_models.rs` on `main` (the enum has since moved out of `core/src/config_types.rs`):

```rust
pub enum ReasoningEffort {
    None, Minimal, Low, Medium, High, XHigh, Max, Ultra,
    /// A model-defined effort value that this client does not know yet.
    Custom(String),
}
```

`Deserialize` is hand-written and delegates to `FromStr`, whose final arm is `effort => Ok(Self::Custom(effort.to_string()))`; the only rejected input is `""` (`"reasoning_effort must not be empty"`). The same file carries a test named `reasoning_effort_json_schema_is_an_open_string`, and the schema description is *"A non-empty reasoning effort value advertised by the model."* — no `enum` constraint. Worth noting the published config reference still lists only `minimal | low | medium | high | xhigh`, which is narrower than what the code accepts; I went with the source.

So `CodexReasoningEffort` now names all eight (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`, `ultra`) and gains a `custom(String)` case mirroring Codex's own escape hatch. `init(rawValue:)` normalises — a named value never becomes `custom`, and `""` returns `nil`. The picker lists the eight named values plus, when the config holds something else, that literal value shown as `"<value> (custom)"` and preselected, so opening the sheet and saving without touching the picker is a no-op. A config with no top-level key still resolves to `high`, exactly as before. New strings are localized in en/fr/vi/zh-Hans.

I also escape the effort through `escapeTOMLString` when building the managed TOML, since a `custom` value is now user-supplied text rather than a fixed literal.

**Destructive scenarios now covered by tests** (`QuotioTests/CodexReasoningEffortTests.swift`, 22 cases)

- top-level value alongside two profile values → top-level wins
- profile-only value, no top-level key → reads `nil`, save writes the `high` default and leaves the profile untouched
- `none` / `max` / `ultra` / a custom `turbo` → open→save round-trips the literal value, verified both by string match and by re-parsing the merged output
- key inside a comment, inside a multi-line string, with a trailing comment, as a quoted key, as a literal string, with an empty value, and with lookalike key names
- `[[array-of-tables]]` also ends the top-level region
- multi-line string bodies survive a save unmodified

The `openThenSave` helper walks the actual sheet path (`parse → build managed TOML → mergeCodexConfig`) rather than asserting on the parser alone, so the round-trip is proven end to end. Full suite: **91 tests, 0 failures**, and the Debug build succeeds.

**Two notes**

- While tracing the write path I found the merge could strip a line inside a user's multi-line string if it happened to look like a managed top-level key (e.g. `model = "…"` inside a `notify = """…"""` block). Since this PR is about data preservation I made `filterExistingCodexLines` and `composeMergedCodexConfig` use the same multi-line-string scanner; it is ~10 lines and has a regression test. Happy to split it out if you would rather keep it separate.
- `base_url` and `model =` in `readCodexConfig` are still matched without section tracking. I left them alone deliberately to keep this diff scoped and because changing `model` detection affects the proxy-detection flow — happy to follow up in a separate PR if you want that tightened too.

Re conflicts: hunks here sit in `readCodexConfig` (~157–180), `buildManagedCodexTOML` (~350) and the merge helpers (~390–640). #467 touches ~655+, ~1005+ and ~1169+, and #477 touches `readOpenCodeConfig` (~215) and ~752+, so all three should stay disjoint in `AgentConfigurationService.swift`.

nguyenphutrong (approved):
The follow-up fully resolves the data-preservation findings. Parsing now reads only the top-level TOML key, and named or custom reasoning-effort values are preserved instead of being replaced with high. Regression coverage includes profile values and custom round trips; the synthetic merge and full suite pass.

PR #475, opened 2026-08-12
fix(menubar): render placeholder instead of 101% for no-data quotas

### LINKED ISSUE #219: The actual usage of Codex is 0, but it shows 101% in the menu bar.
### What happened?

as above

### Steps to reproduce

Image

### Expected behavior

should show 0% in menu bar

### Severity

🟡 Medium - Bug with workaround available

### Affected component

Menu Bar

### Provider (if applicable)

OpenAI Codex

### CLI Agent (if applicable)

None

### macOS Version

Tahoe Version 26.2

### Quotio Version

0.7.6

### App Mode

Full Mode (Proxy running)

### Screenshots

_No response_

### Logs (optional)

```shell

```

### Additional context

_No response_

giaBaoJS:
## Problem

On app open, menu bar quota values briefly show **101%** and, for OpenAI Codex, can stay at 101% even though actual usage is 0 (see the screenshot in #219: the status item shows `101%` while the dropdown for the same Codex account shows Session 0% / Weekly 0%).

## Root cause

Quota values use `remaining = -1` as a "no data yet" sentinel:

- `QuotioApp.quotaItems` initializes `displayPercent = -1` and keeps it when the selected menu bar item's account key cannot be resolved to a fetched quota entry (the Codex case: its quota keys are filename-based, so a stale selection key never matches).
- `MenuBarSettingsManager.totalUsagePercent` returns `-1` when no valid model percentages exist.
- Some fetchers emit `-1` placeholder models (e.g. Trae's `trae-usage` when an account has no active quota buckets — and IDE quotas are persisted in `UserDefaults` and only refreshed by an explicit scan, so the value survives restarts).

In the default "used" display mode, `QuotaDisplayMode.displayValue(from:)` converted that sentinel with `100 - (-1) = 101`, fabricating the 101%. #275/#277 hid the value in the status item view, but the conversion helper still fabricates 101, and the dropdown menu views (`PercentageBadge`, `LowestBarLayout`, `CardGridLayout`, `RingGridLayout`, `MenuModelDetailView` via `menuDisplayPercent`) render it unguarded — a `-1` model still shows as **101%** there today (reproducible with a Trae account that has no active quota buckets, and with any provider whose persisted snapshot carries `-1` while its refresh keeps failing, since `MonitorRefreshCoordinator.refresh` returns the previous snapshot on failure).

## Fix

Handle the sentinel at the display boundary, mirroring the existing `UsageRowV2.isUnknown` pattern from the main window:

- `QuotaDisplayMode.displayValue(from:)` now propagates negative input as `-1` (no data) instead of computing `100 - (-1) = 101`, and clamps valid input to 0–100. This is the single choke point used by the status item, the dropdown, and quota cards, so a fake >100% can no longer be fabricated anywhere.
- Dropdown menu views render `—` in a neutral color for no-data models instead of a fake percentage (`menuPercentText` helper; `menuStatusColor` returns `.secondary` for unknown values so no-data is not painted red).
- `QuotaCard.QuotaSection` gets the same `—` placeholder and a clamped progress width.

No fetcher behavior changes; `-1` remains the internal "unknown" sentinel and now consistently renders as a placeholder.

## Tests

- New `QuotioTests/QuotaDisplayModeTests.swift`:
  - `displayValue(from: -1)` in `.used` mode returned `101` before this change; now returns `-1` (regression test for the reported symptom).
  - Sentinel propagation in both display modes, normal conversions, and out-of-range clamping.
- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` succeeds.
- `xcodebuild test -scheme Quotio -destination 'platform=macOS'` passes (all suites).

Fixes #219

### DIFF
--- a/Quotio/Models/MenuBarSettings.swift
+++ b/Quotio/Models/MenuBarSettings.swift
@@ -161,11 +161,16 @@ enum QuotaDisplayMode: String, Codable, CaseIterable, Identifiable {
         }
     }
     
-    /// Convert a remaining percentage to the display value based on mode
+    /// Convert a remaining percentage to the display value based on mode.
+    /// A negative input is the "no data" sentinel and is propagated as -1 so it can
+    /// never surface as a fake percentage (e.g. 100 - (-1) = 101%). Valid input is
+    /// clamped to 0-100 before conversion.
     func displayValue(from remainingPercent: Double) -> Double {
+        guard remainingPercent >= 0 else { return -1 }
+        let clamped = min(100, max(0, remainingPercent))
         switch self {
-        case .used: return 100 - remainingPercent
-        case .remaining: return remainingPercent
+        case .used: return 100 - clamped
+        case .remaining: return clamped
         }
     }
     
--- a/Quotio/Services/StatusBarMenuBuilder.swift
+++ b/Quotio/Services/StatusBarMenuBuilder.swift
@@ -2274,15 +2274,7 @@ private struct RingGridLayout: View {
         LazyVGrid(columns: columns, spacing: 10) {
             ForEach(models, id: \.name) { (model: ModelBadgeData) in
                 VStack(spacing: 4) {
-                    ZStack {
-                        RingProgressView(percent: menuDisplayPercent(remainingPercent: model.percentage, displayMode: displayMode), size: ringSize, lineWidth: 4, tint: menuStatusColor(remainingPercent: model.percentage, displayMode: displayMode), showLabel: model.percentage >= 0)
-
-                        if model.percentage < 0 {
-                            Text("—")
-                                .font(.system(size: ringSize * 0.24, weight: .bold))
-                                .foregroundStyle(.secondary)
-                        }
-                    }
+                    RingProgressView(percent: menuDisplayPercent(remainingPercent: model.percentage, displayMode: displayMode), size: ringSize, lineWidth: 4, tint: menuStatusColor(remainingPercent: model.percentage, displayMode: displayMode), showLabel: true)
 
                     Text(model.name)
                         .font(.system(size: 10, weight: .medium, design: .rounded))
@@ -2466,7 +2458,17 @@ private struct MenuModelDetailView: View {
             }
 
             if !model.isStandaloneMetric && displayStyle == .ring {
-                RingProgressView(percent: displayPercent, size: 14, lineWidth: 2, tint: statusColor)
+                if RingProgressView.isUnknown(displayPercent) {
+                    // A 14pt ring has no room for a label, so replace it with the
+                    // same placeholder the other display styles render.
+                    Text("—")
+                        .font(.system(size: 10, weight: .semibold, design: .rounded))
+                        .foregroundStyle(statusColor)
+                        .accessibilityLabel("usage.ring".localized())
+                        .accessibilityValue("quota.noDataYet".localized())
+                } else {
+                    RingProgressView(percent: displayPercent, size: 14, lineWidth: 2, tint: statusColor)
+                }
             }
         }
         .padding(.horizontal, 12)
--- a/Quotio/Services/StatusBarMenuBuilder.swift
+++ b/Quotio/Services/StatusBarMenuBuilder.swift
@@ -2145,7 +2145,15 @@ private func menuDisplayPercent(remainingPercent: Double, displayMode: QuotaDisp
     displayMode.displayValue(from: remainingPercent)
 }
 
+/// Formatted percentage for menu rows. A negative remaining percentage means
+/// "no data yet" and renders as a placeholder instead of a fake value like 101%.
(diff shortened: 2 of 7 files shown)


nguyenphutrong (changes requested):
Thank you for fixing the `-1` sentinel handling. I tested this PR merged into the current `master`: the Debug build succeeds, all six new regression tests pass, and the one unrelated concurrency test that failed in the full run passed on an isolated retry (76/77 in the full run, then 1/1 on retry).

There is one remaining no-data path to fix before merging: Ring display still exposes an unknown quota as a real 0% value.

- `MenuModelDetailView` passes `displayPercent == -1` directly to `RingProgressView`. That component clamps the value to zero, so an unknown Antigravity model in the submenu renders as an empty 0% ring and VoiceOver announces “0 percent,” rather than indicating that data is unavailable.
- `RingGridLayout` visually overlays an em dash, but its underlying `RingProgressView` still exposes the same “0 percent” accessibility value, creating conflicting visual and spoken states.

Please make the ring paths follow the same unknown-value contract as the text/card paths—for example, replace or hide the ring when the value is unknown and expose an appropriate no-data accessibility value (or hide the decorative ring from accessibility). This will make the fix consistent across every display style as intended.

The sentinel propagation, neutral coloring, clamping, and other text/card placeholders otherwise look correct.

giaBaoJS:
Thanks for the detailed review — pushed 56d68cd addressing the ring paths.

**Where the contract now lives**

`RingProgressView` itself owns the unknown-value contract, so every consumer gets it (I checked all four call sites):

- Negative `percent` (the `-1` sentinel) skips the progress arc entirely instead of clamping to zero, renders the same `—` placeholder the text/card paths use, and exposes `quota.noDataYet` ("No usage data available") as its accessibility value instead of "0 percent".
- Added `.accessibilityElement(children: .ignore)` so the ring is one element and the placeholder glyph is not announced separately from the value.

**Per call site**

1. `MenuModelDetailView` (`StatusBarMenuBuilder.swift:2460`) — the ring here is 14pt with no label, so an empty track would still look like a real 0%. For an unknown value it now renders the `—` placeholder (same as the non-ring styles in that view) carrying the `usage.ring` label and the no-data accessibility value; a known value renders the ring as before.
2. `RingGridLayout` (`StatusBarMenuBuilder.swift:2277`) — dropped the `ZStack` em-dash overlay. That overlay was the source of the visual/spoken contradiction; the component now renders the placeholder and the matching accessibility value together, so they cannot diverge again.
3. + 4. `AntigravityRingLayout` / `StandardRingLayout` (`QuotaScreen.swift:1317`, `:1450`) — these are ring paths too and had a worse variant of the same bug: `QuotaDisplayHelper.displayPercent` clamps before converting, so `-1` became `100 - 0 = 100`, i.e. a **full red 100% ring** in "used" mode for a model with no data (reachable — several fetchers emit `ModelQuota(percentage: -1)`, e.g. `TraeQuotaFetcher.swift:352`, `FactoryDroidQuotaFetcher.swift:187`). They now use a new `QuotaDisplayHelper.ringPercent(remainingPercent:)` that propagates the sentinel. The existing `displayPercent` is untouched, so the text/bar paths you said look correct keep their current clamping.

**Localization**

No new strings: `quota.noDataYet` already exists and is translated for en, fr, vi and zh-Hans (it is what `QuotaScreen` already shows for the empty state), so `Localizable.xcstrings` is unchanged.

**Tests** — `QuotioTests/RingProgressViewTests.swift` (9 cases) against the decision functions the view body actually calls:

- `isUnknown` / `labelText` / `accessibilityValueText` for the sentinel, real values, and out-of-range input.
- `testSentinelFromDisplayModeReachesRingAsUnknown` pins the real pipeline: `QuotaDisplayMode.used.displayValue(from: -1)` still reads as unknown at the ring.
- `testUnknownAccessibilityValueDiffersFromZeroPercent` encodes your point directly — the unknown announcement must not equal the one for a real 0%.

**Verification**

- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` — succeeds.
- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — `** TEST SUCCEEDED **`, full run, including `MonitorRuntimeTests.testCoordinatorCoalescesConcurrentRefreshForProvider`.
- Rendered-output check (local, not committed, to avoid adding a snapshot test to the suite): rendered `RingProgressView` via `ImageRenderer` at percent `-1`, `0` and `100` — the unknown render is byte-different from both, confirming an unknown ring no longer looks like a real 0%. Same run asserted the announced values: unknown → `"No usage data available"`, 0 → `"0 percent"`.

nguyenphutrong (approved):
The remaining ring-display blocker is resolved across all ring call sites. Unknown values no longer render or announce a real 0%/100% value, and accessibility uses the localized no-data state. The compact menu path uses an accessible placeholder; the synthetic merge and targeted display/ring tests pass.

PR #474, opened 2026-08-12
feat(menubar): add quota bucket selector for menu bar value

giaBaoJS:
Closes #393.

## What this does

Each menu bar item picks what it shows, instead of one global mode for every account:

```swift
nonisolated enum MenuBarBucketSelection: Codable, Hashable, Sendable {
    case auto              // unchanged aggregate, per the Usage Display settings
    case window(QuotaWindow)
    case bucket(String)    // exact ModelQuota.name
}
```

The choice is stored on `MenuBarQuotaItem` and defaults to `nil`, which reads as `.auto`, so existing selections keep their current value. The picker lives in Settings → Menu Bar, one row per selected item; new strings are localized in en/fr/vi/zh-Hans.

Options come from the buckets the provider actually reports for that account, so the cases #393 asks for are reachable: `factory-standard-five-hour` vs `factory-core-five-hour`, `antigravity-gemini-session` vs `antigravity-claude-gpt-session`, and paid buckets such as Cursor `on-demand`.

## Window classification

Windows are declared by the producing fetcher via `ModelQuota.window`, set from the payload field the bucket came from, not matched against the bucket name. This is the shared foundation commit also present in #485.

Name matching would be wrong for shapes shipping today: Codex names the Spark five-hour window `codex-spark`, and Copilot names its monthly quotas `copilot-chat` / `copilot-completions`. Both are covered by regression tests built from real payloads.

Nine producers state no window (Cursor exposes only a `billingCycleEnd` boundary, Warp only a `nextRefreshTime`, Trae only a `resetTime`, Kiro and OpenRouter neither). Those stay unclassified rather than guessed, so those accounts get Automatic plus their individual buckets.

## Behaviour details

- a selection an account no longer offers falls back to the aggregate total rather than showing nothing
- buckets that report no percentage (status rows, currency balances) are not offered, since the menu bar renders a percentage
- `MenuBarQuotaItem` identity stays provider + account key; the bucket choice is an attribute, so adding the same account twice is still rejected

## Verification

```
xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build   ** BUILD SUCCEEDED **
xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'  ** TEST SUCCEEDED **  (98 tests)
```

[...]

PR #476, opened 2026-08-12
feat: add complete app reset in Settings

### LINKED ISSUE #373: How do I completely reset Quotio?
### Your Question

My app got into bugged state (ghost account that no longer exist but will not go away, app only show Github Copilot in the toolbar despite having OpenAI too) and I want to completely erase Quotio and restart clean, no matter what I do (erase all files related to Quotio) the exact same state is back

### Topic

Settings / Configuration

### What have you tried?

Used Pearcleaner to complete erase and reinstall quotio and restarted my computer.

### Additional Context

Mac OS 26
0.14.0 Build 49
Installed with Brew

Shown application showing accounts despite already erased settings.
Image

### Quotio Version (optional)

0.14.0

giaBaoJS:
## Motivation

#373 reports that Quotio got into a bugged state (a ghost account that no longer exists, menu bar missing providers) and that **no amount of uninstalling brings it back to a clean state** — even Pearcleaner plus a reboot. That is expected with the current storage layout: Quotio keeps state in places uninstallers do not reach, most notably **Keychain records** (`dev.quotio.desktop.*` services, plus legacy `proseek.io.vn.Quotio.*` / `com.quotio.*` names), and also `~/Library/Application Support/Quotio` (Monitor account/quota snapshots, proxy binary, config) and the UserDefaults domain (including `persisted.ideQuotas`). There was no in-app way to start fresh.

This PR adds a real "Reset Quotio" capability.

## What it does

A new **Danger Zone** section at the bottom of Settings with a destructive **Reset Quotio…** button. It shows a confirmation alert that lists exactly what will be deleted and what is kept; confirming runs `AppResetService.performFullReset`, which:

1. Stops background work first: atom feed polling, refresh task, request tracker, proxy, tunnel, then orphaned proxy processes (`terminateProxyOnShutdown`, run before the defaults wipe because it reads `proxyPort`/`useBridgeMode`).
2. Wipes every Quotio-owned store, each step best-effort with logging:
   - the whole **UserDefaults persistent domain** (`removePersistentDomain`), so every key ever written is cleared without enumeration — settings, onboarding flags, `persisted.ideQuotas`, menu bar selection, …
   - all **Quotio-owned keychain services** via a new `KeychainHelper.deleteAllItems(service:)` that deletes by service regardless of account name (management keys are per config ID and Monitor credentials per provider account, so account names are dynamic). Legacy service names from earlier releases are included.
   - `~/Library/Application Support/Quotio` (proxy binaries, `config.yaml`, Monitor `accounts-v1.json` / `snapshots-v1.json`, request history) and the app's caches directory.
3. Relaunches the app via `open -n` and exits immediately with `exit(0)`, deliberately skipping normal termination handlers so no in-memory state is written back into the stores that were just wiped. The new instance starts at clean onboarding.

## What is deliberately NOT deleted, and why

- **Provider auth files in `~/.cli-proxy-api`** — these are working credentials usable by CLIProxyAPI directly; deleting them would sign the user out of every provider without warning. Kept, and the confirmation alert says so explicitly.
- **Credentials owned by other tools** (`~/.codex`, gh, Claude CLI, …) — Quotio only reads these; they were never Quotio's to delete.
- **Local IDE databases** (Cursor / Trae / Antigravity) — user-owned application data Quotio only scans.
- **Backups of the user's CLI agent configuration files** — they live in the user's own dotfile directories and exist precisely to survive Quotio's actions.

## Testability

The reset targets are a declarative inventory (`AppResetService.Inventory`): defaults domain + keychain services (sourced from a new `KeychainHelper.quotioOwnedServices` single source of truth) + state directories. `AppResetServiceTests` asserts:

- the inventory covers all known keychain services (current + legacy), the app's defaults domain, Application Support/Quotio and caches — and never reaches into `~/.cli-proxy-api` or other tools' credential paths, so a future store must be registered or the test fails;
- `wipe` clears all known persisted defaults keys (seeded into an isolated suite), removes state directories while leaving user-owned siblings untouched, and requests deletion of every keychain service;
- `deleteAllItems(service:)` removes every account under a service in a real keychain round-trip (unique per-run service name).

## Localization

All new strings are localized in en, fr, vi, and zh-Hans in `Localizable.xcstrings`.

## Verification

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED** (all existing tests plus 6 new `AppResetServiceTests` pass)

Note: this complements #472, which closes one specific resurrection path (auto re-import of deleted Cursor/Trae IDE quotas); this PR gives users the full clean-slate escape hatch #373 asks for. No overlapping edits.

Closes #373


nguyenphutrong (changes requested):
Thank you for adding a much-needed recovery path and for clearly preserving provider-owned credentials. I tested this PR merged into the current `master`: the Debug build succeeds and all six new reset tests pass. The full suite had one unrelated concurrency-test failure, which passed on an isolated retry.

The current implementation is not yet a complete fresh-install reset, and because this is destructive I think these gaps need to be addressed before merging:

1. **Two Quotio-owned stores are missing from the inventory.** Quotio writes per-account device profiles to `~/.quotio/antigravity-profiles`, so those account-linked artifacts survive the reset. PostHog 3.64.1 also stores the anonymous/device identity, configuration, and pending event queues under `~/Library/Application Support/<bundle-id>/<project-token>`, not under `Application Support/Quotio`; that directory survives too. Telemetry should be stopped before deletion, and both owned locations should be included without touching the external Antigravity database or other tools' credentials.

2. **Launch at Login survives the reset.** `SMAppService.mainApp` registration is system state, not UserDefaults. If it was enabled, the supposedly fresh install remains registered to launch at login. Please unregister it as part of reset (and handle/report an unregistration failure).

3. **The app always exits and relaunches even when cleanup fails.** Filesystem errors are only logged, and keychain deletion returns no result; its defensive loop can also stop after 128 successful deletions without verifying that the service is empty. This can reproduce the original “reset did not reset” problem while presenting the new instance as clean. Please return/aggregate failures and keep the current app open with a localized actionable error (or retry path) instead of unconditionally relaunching on an incomplete wipe.

4. **Reset needs a dedicated async quiesce path.** `stopProxy()` is normal runtime behavior: it restarts the warmup scheduler and launches some cleanup tasks without awaiting them. Other in-flight monitor/refresh writers and telemetry queues are not drained. Before removing stores and launching a second process, cancel and await all Quotio writers so deleted state cannot be rewritten during the handoff.

The current inventory tests compare against the same incomplete inventory and hard-code only the two implemented directories, so they cannot detect these omissions. Please add coverage for the Antigravity/PostHog locations, launch-item cleanup, failure propagation, and the completed shutdown ordering.

giaBaoJS:
Thanks for the detailed review — all four points are addressed in `daa339c`. I re-derived the inventory from scratch rather than extending the previous list, since the review proved it was incomplete.

## Owned-vs-external inventory

I enumerated every persistence site in the app (UserDefaults, keychain services, and every `appendingPathComponent` / `urls(for:)` / `homeDirectoryForCurrentUser` path construction), plus the third-party SDKs (PostHog, Sparkle) and the system-state registrations.

### Quotio-owned — removed by the reset

| Store | Location | What lives there | Status |
|---|---|---|---|
| UserDefaults domain | `dev.quotio.desktop` | every key the app writes; cleared with `removePersistentDomain` so no enumeration can go stale | already covered |
| Keychain (current) | `…remote-management`, `…local-management`, `…warp`, `…monitor-auth` | management keys (per config id), Warp tokens, Monitor OAuth credentials | already covered |
| Keychain (legacy) | `proseek.io.vn.Quotio.*` and `com.quotio.*` for remote/local/warp | pre-rename records on upgraded installs | already covered |
| App Support | `~/Library/Application Support/Quotio` | `CLIProxyAPI`, `proxy/`, `config.yaml`, `request-history.json`, `Monitor/accounts-v1.json`, `Monitor/snapshots-v1.json`, `Monitor/antigravity-shadow-v1.json` | already covered |
| Caches | `~/Library/Caches/dev.quotio.desktop` | app caches | already covered |
| **Telemetry (PostHog 3.64.1)** | `~/Library/Application Support/dev.quotio.desktop` | `<project-token>/posthog.anonymousId`, `posthog.distinctId`, `posthog.deviceId`, `posthog.remoteConfig`, `posthog.optOut`, the queue folders; plus pre-3.48 files the SDK migrates from the bundle-id directory itself (`PostHogStorage.getAppFolderUrl` → `applicationSupportDirectoryURL() + projectToken`). No other Quotio path is under the bundle-id directory — everything else is under `Application Support/Quotio` — so the whole directory is telemetry state we own | **added** |
| **Antigravity device profiles** | `~/.quotio/antigravity-profiles` | per-account device fingerprints written by `AntigravityDeviceManager` (`profileStorageDir`, `<email>.json`), account-linked | **added** |
| **HTTP storage** | `~/Library/HTTPStorages/dev.quotio.desktop` and `dev.quotio.desktop.binarycookies` | URLSession cookie/credential store keyed by our bundle id | **added** |
| **Launch at Login** | `SMAppService.mainApp` | system registration, not UserDefaults — survived the reset entirely | **added** |

Only the `antigravity-profiles` subdirectory of `~/.quotio` is removed, never `~/.quotio` itself; a test asserts that.

### External — never touched

| Owner | Location | Why it stays |
|---|---|---|
| CLIProxyAPI | `~/.cli-proxy-api` | provider auth files usable without Quotio |
| Codex CLI / Claude Code / Gemini / Factory | `~/.codex`, `~/.claude`, `~/.gemini`, `~/.factory` | other tools' credentials |
| Agent config backups | `<agent config>.backup.<ts>`, written next to the agent's own config | user-owned |
| Antigravity IDE | `~/Library/Application Support/Antigravity/**` (`state.vscdb`, `storage.json`, WAL/SHM, `.quotio.backup`) | external database Quotio only reads/patches |
| Cursor / Trae / other IDEs | their own Application Support directories | read-only quota scan sources |
| Sparkle | its keys live in the app's UserDefaults domain; no separate store | covered by the domain removal, nothing extra to delete |

## Enforcement

The on-disk stores are now a closed `AppResetService.StoreID` enum, and `testInventoryBuildsEveryDeclaredStoreExactlyOnce` asserts a one-to-one correspondence between the enum cases and what `Inventory.current` builds — adding a case without building it, or building a location without declaring it, fails. `testInventoryPathsMatchTheDocumentedLayout` pins each path, `testWipeRemovesEveryRegisteredStoreAndKeepsExternalData` populates every registered store (including the PostHog token subdirectory, its legacy files, and a device profile) and asserts the adjacent Antigravity/Cursor databases and the CLI credential files survive, and `testInventoryNeverTargetsExternalToolData` asserts no inventory path can point at an external tool. `Inventory.current` now takes injected `libraryURL`/`homeURL`/`bundleID`, so no test ever runs against real Quotio state.

## Teardown ordering

`performFullReset` now quiesces before it deletes, and deletes before it hands off:

1. `TelemetryService.shutdownForReset()` — `optOut` → `reset` → `close`, so PostHog's flush timer cannot re-create its directory after the wipe.
2. `AtomFeedUpdateService.stopPolling()`.
3. `QuotaViewModel.quiesceForReset()` — a dedicated path, not `stopProxy()`. It latches `isQuiescedForReset` (so `restartAutoRefresh`, `restartWarmupScheduler` and the `UserDefaults.didChangeNotification` observer all become no-ops — the domain removal itself fires that notification), then cancels **and awaits** `refreshTask` and `warmupTask`, stops the proxy, awaits the management-client `invalidate()`, stops `RequestTracker` and drains its serial file queue via `drainPendingWrites()`, and awaits in-flight Monitor refreshes via `MonitorRefreshCoordinator.quiesce()`. `stopProxy()` is deliberately not reused: it restarts the warmup scheduler and leaves the tunnel stop and session invalidation as un-awaited detached tasks.
4. `await tunnelManager.stopTunnel()`.
5. `CLIProxyManager.terminateProxyOnShutdown()` — reads `proxyPort`/`useBridgeMode`, so it must run before the domain removal.
6. `wipe(inventory:)` — defaults, keychain, files, then the login item.
7. Relaunch only if the report is empty.

On #484: the relaunch keeps using `exit(0)`, which bypasses `applicationWillTerminate` and therefore the new opt-in quit-time agent-config restore. That is intentional in both directions — a reset must not rewrite the user's own `~/.claude` / `~/.codex` configs, which are outside the inventory, and the restore would otherwise read `restoreAgentConfigsOnQuit` from a domain that was just removed and silently no-op anyway. I've documented this on `relaunch()`.

## Failure propagation

`KeychainHelper.deleteAllItems` now returns a `ServiceWipeResult` and always verifies with `SecItemCopyMatching` that the service is empty afterwards, so neither a silent `SecItemDelete` failure nor falling out of the defensive loop after 128 deletions can be mistaken for a completed wipe. `wipe` aggregates every keychain, filesystem and login-item failure into a `ResetReport`, and each target is still attempted independently so one failure does not skip the rest. `performFullReset` returns that report and only relaunches when it is empty; otherwise the current app stays open and the Settings alert shows a localized message listing exactly what is still on the machine, with a retry button. Covered by `testKeychainFailureIsReportedAndDoesNotStopRemainingStores`, `testUndeletableDirectoryIsReported`, `testLaunchAtLoginUnregistrationFailureIsReported` and `testWipeUnregistersLaunchAtLogin`.

## Localization

`settings.reset.failedTitle` and `settings.reset.failedMessage` added in en/fr/vi/zh-Hans. `settings.reset.confirmMessage` and `settings.reset.footer` updated in all four languages so the confirmation stays an accurate warning — they now name the request history, the anonymous analytics identity and the Antigravity device profiles, state that Launch at Login is turned off, and explicitly say the Antigravity app's own database is kept.

## Verification

`xcodebuild build` and `xcodebuild test -destination 'platform=macOS'` both succeed; all 14 reset tests pass. Trial-merged into current `master` and against #474, #481, #484 and #487 — all four merge cleanly, and the build passes on the `master`-merged tree. #487 also touches `QuotaViewModel`, but only inside the same `refreshTask` that `quiesceForReset` cancels and awaits, so the two compose.

nguyenphutrong (changes requested):
The inventory, login-item cleanup, keychain verification, failure reporting, and most writer draining are now addressed. One lifecycle blocker remains: quiesceForReset() calls proxyManager.stop(), whose process/port cleanup runs in an unstructured detached task and returns immediately. Reset can wipe/relaunch while that cleanup is still active, and the cleanup kills by port, so it can race the replacement proxy. Please await stopAndWait() in the reset quiesce path and add coverage proving process/port cleanup completes before wipe/relaunch. The synthetic merge builds and the full suite passes.

nguyenphutrong:
Thank you for the contribution. This branch now conflicts with master, and the full-reset flow in this PR no longer matches the current product plan. Because bringing it forward would require redesigning the behavior and scope, this PR is being closed. The linked issue will remain open for a future solution.

PR #477, opened 2026-08-12
fix(opencode): preserve user settings when auto-configuring opencode.json

### LINKED ISSUE #176: auto configure opencode will wipe out the current setting
### What happened?

version 0.7.4, in the default opencode setting, I have a field like 
`  "plugin" : [
    "opencode-antigravity-auth@1.2.8",
    // "oh-my-opencode",
    "opencode-openai-codex-auth"
  ],`
and when i apply auto config in quotio, this filed is gone and my plugin is not loading
My other provider settings are gone as well

### Steps to reproduce

1, go to agent
2, click reconfigure
3, configuration mode - auto
4, apply

### Expected behavior

the current config in the opencode.json should stay the same, quotio should only update the "quotio" field, in provider

### Severity

### Affected component

Dashboard

### Provider (if applicable)

None

### CLI Agent (if applicable)

None

### macOS Version

26.2

### Quotio Version

0.7.4

### App Mode

Full Mode (Proxy running)

### Screenshots

_No response_

### Logs (optional)

```shell

```

### Additional context

_No response_

giaBaoJS:
## Root cause

Applying auto-configure for OpenCode wiped user settings (`plugin`, other `provider` entries) from `~/.config/opencode/opencode.json`.

`generateOpenCodeConfig` does try to merge into the existing file, but it parses it with a strict `JSONSerialization` call wrapped in `try?`:

```swift
if fileManager.fileExists(atPath: configPath),
   let existingData = fileManager.contents(atPath: configPath),
   let parsed = try? JSONSerialization.jsonObject(with: existingData) as? [String: Any] {
    existingConfig = parsed
}
```

OpenCode's config is JSONC — comments and trailing commas are valid (the reporter's file contains `// "oh-my-opencode"`). Strict parsing fails on such a file, the failure is swallowed, `existingConfig` stays `{}`, and the subsequent write replaces the whole file with just `$schema` + `provider.quotio`, destroying everything the user had.

## Fix

A new `Quotio/Services/OpenCodeConfigEditor.swift` performs a **field-scoped, text-preserving edit** instead of reparsing and reserializing the document:

- **Only `provider.quotio` is rewritten.** The editor parses the file structurally, locates the exact character range of the managed member, and splices it. Everything else — `plugin`, `mcp`, other providers, comments, key order, indentation, line endings, a UTF-8 BOM — comes back byte-identical. Reconfiguring twice is idempotent.
- **`$schema` is only added to a brand-new file.** An existing file that omits it is left as the user wrote it.
- **The revert path is symmetric**: it deletes only the `quotio` member (dropping `provider` when that was its last entry), and returns "nothing to do" — leaving the file completely untouched — when there is no quotio entry.
- **The parser is fail-closed.** It never manufactures validity: comment removal substitutes a space so adjacent tokens cannot fuse, unterminated block comments and strings are rejected rather than truncating the input, and missing commas, trailing content after the root, duplicate `provider` keys, a non-object `provider`, a non-object root and non-UTF-8 bytes are all refused.
- **Every splice is verified before it is returned**: the result is reparsed and compared against the value we intended to produce. A mismatch throws, and nothing is written.
- In automatic mode, a refusal surfaces a localized error and leaves `opencode.json` untouched. Manual mode still shows a quotio-only snippet to merge by hand (nothing is written).
- The read/detection path uses the same tolerant parser, so Quotio detects its provider inside a commented config.
- A timestamped backup is still written before any modification, via the existing `backupIfPresent` helper.

## Parse-or-refuse decision table

| Input | Decision |
| --- | --- |
| Strict JSON | parse |
| `//` and `/* */` comments, trailing commas | parse, preserved in place on write |
| Comment markers inside string literals (`"a /* not a comment */ b"`, `"…// still in string"`) | parse — string literals are never scanned for comments |
| Comma/brace characters inside string literals (`"value, }"`) | parse |
| CRLF line endings, UTF-8 BOM | parse; both preserved on write |
| Unterminated block comment (anywhere, including after an otherwise valid document) | **refuse** |
| Nested block comment `/* a /* b */ */` | **refuse** (JSONC has no nesting; the stray `*/` is a syntax error) |
| Unterminated string literal | **refuse** |
| Closed block comment between two tokens (`1/* x */2`) | **refuse** — the comment becomes a space, so the tokens stay separate and the document is invalid |
| Missing comma between members | **refuse** |
| Trailing content after the root value | **refuse** |
| Root is not a JSON object | **refuse** |
| Duplicate `provider` key at the root, or duplicate managed key inside `provider` | **refuse** (ambiguous target) |
| `provider` present but not an object (including `null`) | **refuse** (merging would replace user content) |
| File is not valid UTF-8 | **refuse** |
| Spliced result does not reparse into the expected value | **refuse** |
| Missing, empty, or whitespace-only file | write a fresh config (`$schema` + `provider.quotio`) |

Every "refuse" row means the same thing: `opencode.json` is left exactly as it was, and automatic mode reports a localized error.

## Localization

New user-facing strings use the established `localizedStatic()` / `String(format:)` pattern and are added textually to `Localizable.xcstrings` for `en`, `fr`, `vi`, `zh-Hans`: `agents.opencode.notConfigured`, `agents.opencode.parseFailed` (path + reason placeholders) and `agents.opencode.parseError.*` (one per refusal reason).

## Tests

`QuotioTests/OpenCodeConfigTests.swift` — 27 tests, including:

- **Round-trip fidelity**: on a config with unrelated top-level keys, nested objects, arrays, another provider and three comments, `merge` then `remove` reproduces the original **byte for byte**; a separate test asserts the source above `provider` is unchanged and the comments (including the reporter's commented-out plugin line) are still present.
- Reconfiguring twice is idempotent; reconfiguring an existing entry replaces only its value.
- `$schema` is not injected into an existing file that omits it.
- Insertion when `provider` is absent, when it is `{}`, when the last member already has a trailing comma, and when the last member carries a trailing `//` comment.
- CRLF + BOM preservation.
- One refusal test per row of the table above.

Verification:
- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED**, 93 passed / 0 failed

Fixes #176

### DIFF
--- a/Quotio/Services/AgentConfigurationService.swift
+++ b/Quotio/Services/AgentConfigurationService.swift
@@ -215,10 +215,10 @@ actor AgentConfigurationService {
         
         guard fileManager.fileExists(atPath: configPath),
               let data = fileManager.contents(atPath: configPath),
-              let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
+              let json = try? Self.parseOpenCodeJSONObject(data) else {
             return nil
         }
-        
+
         // Check for quotio provider
         guard let providers = json["provider"] as? [String: Any],
               let quotioProvider = providers["quotio"] as? [String: Any],
@@ -752,21 +752,26 @@ actor AgentConfigurationService {
         if mode == .automatic && fileManager.fileExists(atPath: configPath) {
             do {
                 let data = try Data(contentsOf: URL(fileURLWithPath: configPath))
-                var config = try JSONSerialization.jsonObject(with: data) as? [String: Any] ?? [:]
-                
-                // Create backup
-                let backupPath = "\(configPath).backup.\(Int(Date().timeIntervalSince1970))"
-                try fileManager.copyItem(atPath: configPath, toPath: backupPath)
-                
-                // Remove quotio provider
-                if var providers = config["provider"] as? [String: Any] {
-                    providers.removeValue(forKey: "quotio")
-                    config["provider"] = providers.isEmpty ? nil : providers
+
+                // Remove only the Quotio-managed provider entry, preserving
+                // every other user setting (#176). When there is nothing to
+                // remove, leave the file completely untouched.
+                guard let updatedData = try Self.openCodeJSONRemovingQuotioProvider(existing: data) else {
+                    return .success(
+                        type: .file,
+                        mode: mode,
+                        configPath: configPath,
+                        authPath: nil,
+                        shellConfig: nil,
+                        rawConfigs: [],
+                        instructions: "No Quotio provider found in opencode.json. OpenCode already uses its default providers.",
+                        modelsConfigured: 0
+                    )
                 }
-                
-                let updatedData = try JSONSerialization.data(withJSONObject: config, options: [.prettyPrinted, .sortedKeys])
+
+                _ = try backupIfPresent(configPath)
                 try updatedData.write(to: URL(fileURLWithPath: configPath))
-                
+
                 return .success(
                     type: .file,
                     mode: mode,
@@ -1207,24 +1212,20 @@ actor AgentConfigurationService {
         ]
 
         do {
-            var existingConfig: [String: Any] = [:]
+            let existingData = fileManager.contents(atPath: configPath)
 
-            if fileManager.fileExists(atPath: configPath),
-               let existingData = fileManager.contents(atPath: configPath),
-               let parsed = try? JSONSerialization.jsonObject(with: existingData) as? [String: Any] {
-                existingConfig = parsed
+            let jsonData: Data
+            do {
+                jsonData = try Self.mergedOpenCodeJSON(existing: existingData, quotioProvider: quotioProvider)
+            } catch {
+                if mode == .automatic && existingData != nil {
+                    // Never fall back to overwriting a file we could not parse:
+                    // that wipes user settings like `plugin` and other providers (#176).
+                    return .failure(error: "Could not parse existing opencode.json at \(configPath): \(error.localizedDescription). The file was left untouched — fix its syntax and try again, or use manual mode.")
+                }
+                jsonData = try Self.mergedOpenCodeJSON(existing: nil, quotioProvider: quotioProvider)
             }
-
-            if existingConfig["$schema"] == nil {
-                existingConfig["$schema"] = "https://opencode.ai/config.json"
-            }
-
-            var providers = existingConfig["provider"] as? [String: Any] ?? [:]
-            providers["quotio"] = quotioProvider
-            existingConfig["provider"] = providers
-
-            let jsonData = try JSONSerialization.data(withJSONObject: existingConfig, options: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes])
-            let jsonString = String(data: jsonData, encoding: .utf8) ?? "{}"
+            let jsonString = String(decoding: jsonData, as: UTF8.self)
 
             let rawConfigs = [
                 RawConfigOutput(
@@ -1239,11 +1240,7 @@ actor AgentConfigurationService {
             if mode == .automatic {
                 try fileManager.createDirectory(atPath: configDir, withIntermediateDirectories: true)
 
-                var backupPath: String? = nil
-                if fileManager.fileExists(atPath: configPath) {
-                    backupPath = "\(configPath).backup.\(Int(Date().timeIntervalSince1970))"
-                    try? fileManager.copyItem(atPath: configPath, toPath: backupPath!)
-                }
+                let backupPath = try backupIfPresent(configPath)
 
                 try jsonData.write(to: URL(fileURLWithPath: configPath))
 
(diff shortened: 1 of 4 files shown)


Quotio/Services/AgentConfigurationService.swift
```diff
@@ -1329,7 +1326,136 @@ actor AgentConfigurationService {
...
+                }
+                if c == "\"" { inString = false }
+                i += 1
+            } else if c == "\"" {
+                inString = true
+                withoutComments.append(c)
+                i += 1
+            } else if c == "/", i + 1 < chars.count, chars[i + 1] == "/" {
+                while i < chars.count, chars[i] != "\n" { i += 1 }
+            } else if c == "/", i + 1 < chars.count, chars[i + 1] == "*" {
```
nguyenphutrong: This pre-pass can make malformed JSONC look valid, defeating the new ‘leave unparseable files untouched’ guarantee. For example, `[{"value": 1/* missing comma */2}]` is transformed so the two numeric tokens join, and an otherwise valid document followed by an unterminated `/*` comment is accepted because the rest of the input is silently discarded. Two temporary regression tests expecting `parseOpenCodeJSONObject` to throw for these inputs both fail. Please preserve a token boundary when removing a closed block comment, reject unterminated block comments, and add coverage for both cases before automatic mode is allowed to rewrite the file.

Quotio/Services/AgentConfigurationService.swift
```diff
@@ -1329,7 +1326,136 @@ actor AgentConfigurationService {
...
+        if let existing {
+            object = try parseOpenCodeJSONObject(existing)
+        }
+        if object["$schema"] == nil {
+            object["$schema"] = "https://opencode.ai/config.json"
+        }
+        var providers = object["provider"] as? [String: Any] ?? [:]
+        providers["quotio"] = quotioProvider
+        object["provider"] = providers
+        return try JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes])
```
nguyenphutrong: Re-serializing the whole object removes every JSONC comment and normalizes all formatting/key order; the helper also adds `$schema` when an existing file lacks it. That conflicts with #176’s expected behavior that the existing config stay the same and Quotio only update `provider.quotio`, and it specifically drops the reporter’s commented plugin line. A backup does not make the active config unchanged. Please use a JSONC-preserving, field-scoped edit so content outside the managed provider entry survives verbatim, and add a regression test that compares the surrounding source before and after the edit.

Quotio/Services/AgentConfigurationService.swift
```diff
@@ -1207,24 +1212,20 @@ actor AgentConfigurationService {
...
-            if existingConfig["$schema"] == nil {
-                existingConfig["$schema"] = "https://opencode.ai/config.json"
+            let jsonData: Data
+            do {
+                jsonData = try Self.mergedOpenCodeJSON(existing: existingData, quotioProvider: quotioProvider)
+            } catch {
+                if mode == .automatic && existingData != nil {
+                    // Never fall back to overwriting a file we could not parse:
+                    // that wipes user settings like `plugin` and other providers (#176).
+                    return .failure(error: "Could not parse existing opencode.json at \(configPath): \(error.localizedDescription). The file was left untouched — fix its syntax and try again, or use manual mode.")
```
nguyenphutrong: This new error, and the new `No Quotio provider found…` instruction in the default-config path, are displayed to users as hardcoded English. The repository rule requires user-facing strings to use the existing localization approach. Please add localization keys (including parameter placeholders for the path/error) and resolve them with the established `localizedStatic()`/formatted-string pattern.

nguyenphutrong (changes requested):
Thank you for addressing the destructive OpenCode merge path and adding focused coverage. The PR merges cleanly into the current `master`; the Debug build and all 83 existing tests pass. Changes are requested because the new parser can still turn malformed JSONC into valid data and overwrite it, the write path removes content outside `provider.quotio`, and the newly introduced user-facing messages bypass localization. Details are inline.

giaBaoJS:
Thanks — all three points are addressed in 0e2f30d. The JSONC helpers moved out of `AgentConfigurationService.swift` into a new `Quotio/Services/OpenCodeConfigEditor.swift`, which shrank the diff in the shared file to call sites only (18 added lines).

**1. The pre-pass could make malformed JSONC look valid** (`AgentConfigurationService.swift:1412`)

Both of your cases now refuse, at two independent layers:

- The stripper replaces each removed comment with a single space, so `[{"value": 1/* missing comma */2}]` becomes `[{"value": 1 2}]` — invalid — instead of `[{"value": 12}]`. `testBlockCommentRemovalPreservesTokenBoundary` asserts the exact stripped text and that `parseObject`/`merging` throw.
- The stripper throws `.unterminatedBlockComment` / `.unterminatedString` instead of consuming to EOF, so a valid document followed by an unterminated `/*` is rejected rather than accepted by discarding the tail. `testUnterminatedBlockCommentIsRejected` covers exactly that shape plus the mid-document one.

Beyond the stripper, every write now goes through a structural JSONC parser that rejects missing commas, trailing content after the root, a non-object root, duplicate `provider` keys, a non-object `provider`, and non-UTF-8 bytes. Finally, the spliced result is reparsed and compared against the value we intended to write; a mismatch throws and nothing is written.

**Parse-or-refuse decision table**

| Input | Decision |
| --- | --- |
| Strict JSON | parse |
| `//` and `/* */` comments, trailing commas | parse, **preserved in place** on write |
| Comment markers inside string literals (`"a /* not a comment */ b"`, `"…// still in string"`) | parse — string literals are never scanned for comments |
| Comma/brace characters inside string literals (`"value, }"`) | parse |
| CRLF line endings, UTF-8 BOM | parse; both preserved on write |
| Unterminated block comment, anywhere including after an otherwise valid document | **refuse** |
| Nested block comment `/* a /* b */ */` | **refuse** — JSONC has no nesting, the stray `*/` is a syntax error |
| Unterminated string literal | **refuse** |
| Closed block comment between two tokens (`1/* x */2`) | **refuse** — becomes `1 2`, so the tokens cannot fuse |
| Missing comma between members | **refuse** |
| Trailing content after the root value | **refuse** |
| Root is not a JSON object | **refuse** |
| Duplicate `provider` key at root, or duplicate managed key inside `provider` | **refuse** — ambiguous target |
| `provider` present but not an object (including `null`) | **refuse** — merging would replace user content |
| File is not valid UTF-8 | **refuse** |
| Spliced result does not reparse into the expected value | **refuse** |
| Missing, empty or whitespace-only file | write a fresh config (`$schema` + `provider.quotio`) |

Every "refuse" row leaves `opencode.json` byte-identical and reports a localized error in automatic mode. There is one test per row.

**2. The write path removed content outside `provider.quotio`** (`AgentConfigurationService.swift:1366`)

You were right, and the fix is a genuine field-scoped edit rather than a reserialization. `OpenCodeConfigEditor` parses the document structurally, takes the character range of the `provider.quotio` member, and splices only that range back into the original text. Consequences:

- Comments — including the reporter's `// "oh-my-opencode"` line — key order, indentation, line endings and a UTF-8 BOM all survive verbatim.
- `$schema` is no longer injected into an existing file that omits it; it is only written for a brand-new file. (`testMergeDoesNotAddSchemaToAnExistingFileThatOmitsIt`)
- The revert path deletes only the `quotio` member — plus one adjacent comma and its own line — and drops `provider` only when that was its last entry. `testRemoveDeletesOnlyQuotioAndKeepsComments` compares the full expected text.

The regression tests you asked for:

- `testMergeThenRemoveRestoresTheOriginalByteForByte` — on a config modelled on #176 (unrelated top-level keys, `plugin` with a commented-out entry, `mcp`, nested objects and arrays, another provider, block and line comments), merge-then-remove returns the original string **byte for byte**.
- `testMergeKeepsEverythingOutsideProviderQuotioVerbatim` — asserts the merged text still starts with the untouched source above `provider`, and still contains each comment and unrelated member.
- `testReconfiguringTwiceIsIdempotent` — merging an already-merged file is a no-op.

**3. Hardcoded English** (`AgentConfigurationService.swift:1224`)

Both new messages now resolve through `localizedStatic()`, with `en`/`fr`/`vi`/`zh-Hans` added textually to `Localizable.xcstrings` (no catalog reserialization):

- `agents.opencode.notConfigured` — the default-config "no Quotio provider" instruction.
- `agents.opencode.parseFailed` — the refusal error, with `%1$@` for the path and `%2$@` for the reason.
- `agents.opencode.parseError.*` — one key per refusal reason, resolved via `OpenCodeConfigError.errorDescription`, so the reason interpolated into `parseFailed` is localized too (including line/column for a syntax error).

I left pre-existing strings in these functions alone, since #482 already localizes them.

**Correction to the PR description**: the previous "comments are not preserved on rewrite" caveat no longer applies and has been removed — comments now survive. The backup is still taken before every write.

**Interaction with the sibling PRs**

- **#482** (`feat/opencode-split-gemini-provider`) rewrites the same two functions. `OpenCodeConfigEditor.merging(existing:providers:)` and `removingProviders(existing:keys:)` already take a *set* of managed provider keys, so #482's split is `["quotio": …, "quotio-gemini": …]` and `keys: ["quotio", "quotio-gemini"]` — no editor change needed, and multi-entry insertion/removal is already handled. Expect textual conflicts in `generateOpenCodeConfig` / `generateOpenCodeDefaultConfig` and in the `agents.opencode.*` block of `Localizable.xcstrings`; my keys sort after all five of #482's (`notConfigured` < `parseError.*` < `parseFailed`, all after `mergeProviders`), so the catalog resolves by concatenation in key order. I did not touch `agents.opencode.default.removed` / `.manual`, which #482 owns.
- **#467** and **#473** touch `AgentConfigurationService.swift` in the Codex/Amp regions and, for #473, the `agents.reasoningEffort.*` catalog block — no overlap with the OpenCode hunks or catalog range here.

**Verification**

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED**, 93 passed / 0 failed (27 in `OpenCodeConfigTests`). `MonitorRuntimeTests.testCoordinatorCoalescesConcurrentRefreshForProvider` flaked on two intermediate parallel runs and passes both in isolation and in the final full run.

nguyenphutrong (approved):
The follow-up resolves all prior blockers. Malformed JSONC now fails closed, field-scoped splices preserve content outside the managed provider, and the new errors/instructions are localized. Focused OpenCode tests pass; the one unrelated full-suite coalescing failure passed immediately in isolation.

PR #479, opened 2026-08-12
feat(trae): add Trae CN detection to Scan for IDEs

### LINKED ISSUE #265: add support for trae cn
### Problem Statement

Now, only can scan ide for trae ,but not scan ide for trae cn.

### Proposed Solution

Image

### Alternative Solutions

_No response_

### Feature Category

New AI Provider Integration

### Use Case

_No response_

### Mockups / Examples

_No response_

### Additional Context

_No response_

### Checklist

giaBaoJS:
## What

"Scan for IDEs" currently detects only the international Trae edition. This PR adds detection for **Trae CN**, the China edition requested in #265.

Trae CN is a separate app, not a locale of Trae — both can be installed side by side:

| | Trae | Trae CN |
|---|---|---|
| App bundle | `Trae.app` | `Trae CN.app` (bundle id `cn.trae.app`) |
| Storage | `~/Library/Application Support/Trae/User/globalStorage/storage.json` | `~/Library/Application Support/Trae CN/User/globalStorage/storage.json` |
| API host fallback | `https://api-sg-central.trae.ai` | `https://api.trae.cn` |
| Web origin | `https://www.trae.ai` | `https://www.trae.com.cn` |

Sources: [Homebrew cask `trae-cn`](https://formulae.brew.sh/cask/trae-cn) (app name, bundle id, zap path) and https://www.trae.com.cn/. Both editions store the same `iCubeAuthInfo://icube.cloudide` auth blob in `storage.json` and use the same entitlement quota API shape; the fetcher already prefers the `host` field from the auth blob, so the CN default host is only a fallback.

## How

- **`TraeQuotaFetcher`** is refactored around a `TraeVariantDescriptor` (provider, display name, app bundle, Application Support directory, default API host, web origin) with `.international` and `.cn` presets — no copy-pasted second fetcher; the same actor is instantiated once per variant.
- **New `AIProvider.traeCn`** (`"trae-cn"`, display name "Trae CN", menu bar symbol `TC`) so both editions coexist with distinct provider identities and stable account keys. It reuses the existing `trae` / `trae-menubar` assets.
- **IDE scan flow**: opt-in "Trae CN IDE" toggle in the scan sheet, privacy-notice path listing, scan result row, scoped refresh (`refreshTraeCnQuotasInternal`), Monitor-mode `localIDE` source classification, and inclusion in `ideProvidersToSave` for UserDefaults persistence.
- **Gating parity with Trae**: `.traeCn` joins `.trae` in `usesBrowserAuth`, `supportsManualAuth == false`, and `isQuotaTrackingOnly`, and `finishScopedRefresh` now keys persistence off `ideProvidersToSave` instead of a hardcoded `cursor || trae` check. This composes with #472 (deletable scanned IDE accounts / auto-refresh exclusion): everything there is gated on `usesBrowserAuth` and `ideProvidersToSave`, so Trae CN automatically gets the same delete + no-resurrection behavior once both are merged.
- **Localization**: new `ideScan.traeCn.detail` string in en/fr/vi/zh-Hans.

## Tests

New `QuotioTests/TraeVariantTests.swift` (8 tests): variant descriptors produce the correct per-edition paths/hosts and never collide, `.traeCn` identity is distinct from `.trae` while mirroring its gating traits, Monitor accounts classify as `localIDE`, and scan options/accessed paths/privacy items distinguish the two editions.

## Verification

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS'` — **TEST SUCCEEDED** (full suite)
- `-only-testing:QuotioTests/TraeVariantTests` — 8/8 passed

Closes #265


Quotio/Services/QuotaFetchers/TraeQuotaFetcher.swift
```diff
@@ -61,32 +119,27 @@ actor TraeQuotaFetcher {
...
         return authData?.accessToken != nil
     }
-    
-    /// Read auth data from Trae's storage.json
+
+    /// Read auth data from this edition's storage.json
     func readAuthFromStorageJson() -> TraeAuthData? {
-        let expandedPath = NSString(string: storageJsonPath).expandingTildeInPath
-        
+        let expandedPath = NSString(string: variant.storageJsonPath).expandingTildeInPath
```
nguyenphutrong: The correct Trae CN file/key will still be reported as not found downstream because current Trae CN clients do not store `iCubeAuthInfo://icube.cloudide` as plaintext JSON. Two independent current implementations decode it as a Base64 envelope (6-byte header + 32-byte key + AES-128-CBC ciphertext) and validate a SHA-512-prefixed payload before parsing JSON: [WorkBuddy-Switch](https://github.com/koi128bit/WorkBuddy-Switch/blob/master/Sources/OpenUsage/TraeSupport.swift#L222-L265) and [trae-mate](https://github.com/luckymiaow/trae-mate/blob/master/src-tauri/src/trae_auth.rs#L23-L100). Please support the current encrypted form (and plaintext only if a real legacy fixture requires it), then add a sanitized representative storage fixture proving the scan extracts the account without exposing credentials.

Quotio/Services/QuotaFetchers/TraeQuotaFetcher.swift
```diff
@@ -139,8 +192,8 @@ actor TraeQuotaFetcher {
             return nil
         }
         
-        // Use the API host from auth data or default
-        let apiHost = authData.apiHost ?? "https://api-sg-central.trae.ai"
+        // Use the API host from auth data or the edition's default
+        let apiHost = authData.apiHost ?? variant.defaultAPIHost
```
nguyenphutrong: Please verify the CN endpoint and response with a sanitized real fixture instead of inheriting the v1 endpoint on the next line. Current public CN integrations use either `/trae/api/v2/pay/ide_user_ent_usage` with a v1 fallback ([source](https://github.com/koi128bit/WorkBuddy-Switch/blob/master/Sources/OpenUsage/TraeUsageService.swift#L305-L326)) or `/trae/api/v2/pay/user_current_entitlement_list` ([source](https://github.com/luckymiaow/trae-mate/blob/master/src-tauri/src/checkin.rs#L343-L410)); neither corroborates v1 `user_current_entitlement_list`. A request/response fixture should cover the chosen endpoint, headers, and CN response shape so a successfully decoded account does not immediately fail quota detection.

Quotio/Models/Models.swift
```diff
@@ -232,7 +235,7 @@ nonisolated enum AIProvider: String, CaseIterable, Codable, Identifiable, Sendab
     /// GLM and ClinePass are excluded because they should only be added via Custom Providers
     var supportsManualAuth: Bool {
         switch self {
-        case .cursor, .trae, .devin, .grok, .glm, .clinePass:
+        case .cursor, .trae, .traeCn, .devin, .grok, .glm, .clinePass:
```
nguyenphutrong, line 238: Adding `.traeCn` to this branch also makes `refreshAutoDetectedProviders()` scan it on every general provider refresh, because that method iterates every provider where `supportsManualAuth == false`. That reads `~/Library/Application Support/Trae CN/.../storage.json` even when the user did not enable the new `scanTraeCn` option, contrary to `IDEScanSettingsManager`’s explicit-consent contract. Please gate local IDE reads on recorded consent/discovery (or otherwise keep Trae CN out of the ungated refresh path) and add a test showing an ordinary refresh cannot access this source before opt-in.

Quotio/Views/Components/IDEScanSheet.swift
```diff
@@ -210,7 +220,14 @@ struct IDEScanSheet: View {
                     } else if scanOptions.scanTrae {
                         resultRow(icon: "xmark.circle.fill", color: .secondary, text: "Trae: " + "ideScan.notFound".localized())
                     }
-                    
+
+                    if result.traeCnFound {
+                        resultRow(icon: "checkmark.circle.fill", color: .green, text: "Trae CN: \((result.traeCnEmail ?? "Found").masked(if: settings.hideSensitiveInfo))")
```
nguyenphutrong: The new `Trae CN IDE`, `Trae CN:`, and `Found` text is user-facing but hardcoded. Repository guidance requires the existing localization approach for user-facing strings; please add keys for the new title/result formatting and use `.localized()` (including the fallback state).

nguyenphutrong (changes requested):
Thank you for the comprehensive provider plumbing. The Trae CN app name and Application Support path are supported by the Homebrew cask, the PR merges cleanly into current `master`, the Debug build succeeds, and all 79 tests pass (including the 8 new tests). I cannot approve yet because the tests cover descriptors rather than the actual scan: the current auth format is not decoded, the CN endpoint assumption is unverified, and normal refresh bypasses the explicit scan-consent toggle. Details are inline.

giaBaoJS:
Thanks — all four points were valid. Pushed in 990f0c2. Here is what is now evidenced, what I narrowed, and how consent is enforced.

**1. Auth format (was: not decoded).** Fixed. `TraeAuthEnvelope` decodes the Base64 envelope `header(6) ‖ randomKey(32) ‖ AES-128-CBC( SHA512(payload) ‖ payload )`, with `key/iv = SHA512( SHA512(randomKey) ‖ secret )[0..<32]`, verifying the SHA-512 prefix before parsing. Plaintext JSON is still accepted first so an older `storage.json` keeps working. This applies to both editions, since they share the code path.

The two implementations you linked agree byte-for-byte, and I verified that rather than assuming it: trae-mate stores the secret obfuscated as two 64-byte halves XORed at runtime, and their XOR equals WorkBuddy-Switch's constant exactly. `TraeCnScanTests` encodes its fixtures with **trae-mate's halves** while production decodes with **WorkBuddy-Switch's constant**, so the cross-check is enforced by the suite, not by a comment.

**2. CN endpoint (was: unverified).** Narrowed rather than guessed. I have no Trae CN install and could not capture a real request/response, so I did not invent one:

- **Host:** the hardcoded `https://api.trae.cn` default is **removed** — `TraeVariantDescriptor.cn.defaultAPIHost` is now `nil`. The host comes from the auth blob's own `host` field, which is what both references do (`TraeUsageService.baseURL(rawHost: payload.host, …)`, `trae-mate`'s `cred.host`). A blob without a host yields no request instead of a guessed endpoint. The international edition keeps its existing `api-sg-central.trae.ai` fallback (in WorkBuddy-Switch's allow-list). Since the host now comes from a user-writable file, it is constrained to https under `trae.ai` / `trae.cn` / `trae.com.cn` so a tampered `storage.json` cannot redirect the bearer token.
- **Path:** CN no longer inherits v1 `user_current_entitlement_list`. It tries `trae/api/v2/pay/ide_user_ent_usage`, then the v1 `ide_user_ent_usage` fallback (both from WorkBuddy-Switch's `TraeUsageService`), then the international v1 entitlement endpoint last so an install answering only there still works. International's list is unchanged.
- **Response:** rather than assume a root key I cannot verify, entitlement packs are found structurally (objects carrying `entitlement_base_info` + `usage`) when `user_entitlement_pack_list` is absent — the same approach as WorkBuddy-Switch's `collectEntitlementPacks`. Tests cover both shapes so a decoded account does not immediately fail quota detection.

Stated plainly: the format, field names and host-from-blob behaviour are evidenced; **which** endpoint a real CN install answers on is not, so it is an ordered candidate list derived from public CN integrations instead of a single assumed path.

**3. Tests cover descriptors, not the scan.** `QuotioTests/TraeCnScanTests.swift` runs the real path: build an envelope → write a real `storage.json` → point the real `TraeQuotaFetcher` at it via path overrides → `readAuthFromStorageJson()` → `parseQuotaResponse` → `makeProviderQuota`, asserting the extracted account (`token`, `refreshToken`, `userId`, `account.email`, `account.username`, `host`) and the resulting quota (`premium-fast` 150/600, 75%, plan `Pro`). Also covered: legacy plaintext still decodes, a tampered payload fails the integrity check, garbage is rejected, an unsafe host is rejected, and an unrecognised response yields no quota.

On provenance, to be explicit: **no real Trae CN credential was captured.** The fixture reproduces the documented/observed international format from the two implementations above; the values are synthetic. The tests therefore assert the decode contract, not a captured sample — I did not want to present a constructed blob as a real one.

**4. Consent bypass.** Fixed. The gate is at the top of `refreshQuota(for:)`, before the monitor and normal branches, so `refreshAutoDetectedProviders()` — which iterates every `supportsManualAuth == false` provider — cannot reach `~/Library/Application Support/Trae CN/…` in either mode. `mayRefreshScanGatedProvider(imported:)` allows a read only when an explicit scan already imported an account; with nothing imported the file is not opened at all. `refreshTraeCnQuotasInternal()` keeps the same guard as defence in depth. This mirrors the rule #486 applies to Cursor/Trae.

`TraeQuotaFetcher.storageReadAttemptCount` makes this directly observable, so the test asserts the *absence of a read* rather than an absent result: before opt-in the count is 0 and no account is imported; after an explicit scan imported one, the same call reads (count 1). That second half is the control that stops the first from passing vacuously. A separate test asserts `.traeCn` really is in the auto-detected sweep, so the gate is guarding a live path.

**5. Localization.** `ideScan.traeCn.title`, `ideScan.traeCn.result` (`%@` format, used for both the found and not-found states) and `ideScan.found` added for en/fr/vi/zh-Hans and used via `.localized()`. No hardcoded user-facing text left in the new UI.

**Verification.** Debug build succeeds; `xcodebuild test -destination 'platform=macOS'` passes — 88 tests, 13 new. I also trial-merged in a throwaway worktree against current `master` plus #472 and #486: `master` merges cleanly, and #486 produces exactly one trivial conflict, the `ideProvidersToSave` line (`[.cursor, .trae]` vs `[.cursor, .trae, .traeCn]`); resolved as the union it builds and all 108 tests pass. Whichever lands second should take the union and add `.traeCn` to #486's `fetchImportedIDEQuotas(for:)` switch so the global refresh updates it too — #486's own gate already keeps that consent-safe.

One thing I deliberately left alone: Cursor and Trae have the same ungated-refresh gap on `master`, but changing their behaviour is #486's territory, so I scoped this to Trae CN rather than altering existing providers here. Happy to widen it if you'd prefer it in one place.

nguyenphutrong (changes requested):
The encrypted auth format, actual storage scan, consent gate, and localization are now covered. One core blocker remains: Trae CN quota transport is still unverified. The implementation tries endpoint candidates derived from third-party integrations plus an international fallback, while tests invoke the parser directly and do not exercise fetchQuota request construction, headers/body, endpoint fallback, or a representative CN response. Please provide evidence from a sanitized real CN capture and add a URL-loading test for the selected host/path, request, fallback, and response shape. The synthetic merge and full suite otherwise pass.

nguyenphutrong:
Thank you for the contribution. This branch now conflicts with master, and the proposed Trae CN integration no longer matches the current provider roadmap and architecture. This PR is therefore being closed rather than asking for a substantial rework. The feature can be reconsidered in a new PR if it is prioritized again.

PR #480, opened 2026-08-12
feat(stats): persist traffic statistics across proxy restarts

### LINKED ISSUE #359: 流量统计清理问题
### Your Question

每次重启 proxy 或者升级 cpa 之后，流量统计都会清零，也没有看到哪里有配置可以保留

### Topic

Settings / Configuration

### What have you tried?

_No response_

### Additional Context

_No response_

### Quotio Version (optional)

_No response_

giaBaoJS:
## Problem

Every proxy restart — including the restart that happens during a CLIProxyAPI upgrade — resets the dashboard traffic statistics (requests, tokens, success rate) to zero, and there is no way to retain them.

## Root cause

The dashboard KPI cards read `QuotaViewModel.usageStats`, which is fetched verbatim from CLIProxyAPI's `GET /v0/management/usage` endpoint (`ManagementAPIClient.fetchUsageStats()`). Those counters live **in memory inside the CLIProxyAPI process**, so any proxy restart or binary upgrade starts them over from zero. Quotio never persisted them.

(`RequestTracker` / `request-history.json` is a separate per-request log capped at 50 entries and does not back these KPIs.)

## Design

`UsageStatsAggregator` accumulates the proxy's counters on the Quotio side and persists them. Correctness rests on an **explicit session boundary driven by the proxy lifecycle**, not on inferring a restart from a counter decrease.

### 1. Session boundary from the managed lifecycle

`CLIProxyManager` owns a `ProxyUsageSessionRecorder` and drives it from every path on which it tears the managed proxy down:

| Lifecycle path | `CLIProxyManager.swift` | Hook |
| --- | --- | --- |
| `stop()` — user toggle, mode switch (`QuotaViewModel.stopProxy()`, `SettingsScreen`) | `stop()` | `proxyDidStop()` (synchronous path, no final fetch possible) |
| `stopAndWait()` — health recovery (`QuotaViewModel.attemptProxyRecovery()`) | `stopAndWait()` | `await proxyWillStop()` |
| Config-change restart — port, host, allow-remote, logging, routing strategy, proxy URL | `restartProxyIfRunning()` | `await proxyWillStop()` then `stop()` |
| **Upgrade** — `performManagedUpgrade(to:)` → promote new binary | `promote(version:source:)` | `await proxyWillStop()` then `stop()` |
| Rollback | `rollback()` | `await proxyWillStop()` then `stop()` |

`proxyWillStop()` takes a **final `/v0/management/usage` snapshot** from the process that is about to go down (the fetch closure is installed by `QuotaViewModel`, which owns the management client), folds it, and closes the session: the session's counters move into the persistent baseline and the live sample is zeroed. The next process may then start from zero — or from any value at all — without discarding what the previous one served. `proxyDidStop()` closes the session without a fetch and is idempotent, so paths that run both still fold exactly once.

### 2. Per-proxy scoping

Persisted state is keyed by `UsageStatsSource`: `.localProxy` for the managed proxy, `.remote(endpoint:)` keyed on the normalized management base URL. Local totals and each remote server's totals are separate buckets that are never added together, so switching modes — or changing the configured remote server — no longer mixes unrelated counters, on screen or on disk. `QuotaViewModel.initialize()` re-scopes the displayed totals on every mode switch. Version 1 files migrate onto the local bucket.

### 3. Counter-decrease fallback

The original "new sample is lower than the last one" check is retained, but **only** as a fallback for restarts Quotio does not perform. It is not what the managed paths depend on.

## Honest limitation

Retention is exact for restarts Quotio performs, and best-effort otherwise:

- **Managed stop / config restart / upgrade / rollback / health recovery** — exact. A final snapshot is folded before the process is signalled, so requests served between the last poll and the shutdown are counted.
- **`stop()` from the synchronous path** — the session is closed correctly, but no final snapshot is fetched, so requests served since the last poll (up to one refresh interval) are not counted.
- **Proxy dies without Quotio's involvement** — crash, external `kill`, machine sleep or power loss, or `terminateProxyOnShutdown()` at app quit. Quotio gets no chance to snapshot, so whatever the proxy served since the last poll is genuinely unrecoverable. The counter-decrease fallback preserves everything up to that last poll; it cannot recover what came after. This PR does not claim perfect retention for these cases.

## Retention / storage behavior

- The persisted file holds **fixed-size aggregate counters only** (no per-request entries), so it never grows with traffic — no rotation needed (a test asserts it stays under 2 KB after 500 samples). One small entry per proxy identity.
- A corrupt/unreadable file is discarded and the aggregator starts fresh, mirroring `RequestTracker`'s recovery behavior.
- Fresh installs behave exactly as before (`restoredStats(for:)` returns nil until traffic exists); the 404 fallback for older CLIProxyAPI versions without the usage endpoint is unchanged. No UI changes.
- `reset()` clears every source's totals (persisted immediately).

## Concurrency

Everything runs on `@MainActor`. Both suspension points are handled explicitly:

- `refreshData()` resolves the `UsageStatsSource` **together with** the `apiClient` it belongs to, before the awaits, so a mode switch mid-refresh cannot file one proxy's counters under another.
- `proxyWillStop()` awaits the final fetch, then re-reads the stored sample and merges component-wise (counters only grow within a session), so a poll that lands during the suspension is neither lost nor double-counted.
- In `stopAndWait()`, `promote()`, `rollback()` and `restartProxyIfRunning()`, the hook is awaited **before** `proxyStatus.running` / process state is captured, never after.

## Tests

`QuotioTests/UsageStatsAggregatorTests.swift` — both reviewer scenarios are covered by name:

- `testManagedRestartFoldsFinalSnapshotTakenAtTeardown` — last poll 100, nine more requests, managed restart, first new sample 1 → **110**
- `testManagedRestartRetainsTotalsWhenNoDecreaseIsEverObserved` — last poll 5, restarted proxy already at 10 by the next poll (no decrease ever observed) → **15**
- `testSessionBoundaryIsIdempotent` — `proxyWillStop()` + repeated `proxyDidStop()` fold exactly once
- `testConcurrentPollDuringFinalSnapshotIsMergedNotDoubleCounted` — refresh landing during the shutdown fetch
- `testLocalAndRemoteTotalsAreNeverMerged` — local 100 → remote 20 shows **20**; back to local 100 shows **100**
- `testDistinctRemoteServersKeepSeparateTotals`, `testMixedSourceTotalsStaySeparateAcrossRelaunch`, `testLifecycleBoundaryOnlyClosesLocalProxySession`, `testRemoteEndpointKeyIsNormalized`
- `testVersion1FileMigratesOntoLocalProxy`
- plus the original coverage: unmanaged-restart fallback, monotonic samples, app relaunch, bounded file, corrupt file, reset

## Verification

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — **TEST SUCCEEDED**, 84 passed / 0 failed

Closes #359

### DIFF
--- a/Quotio/Services/Proxy/CLIProxyManager.swift
+++ b/Quotio/Services/Proxy/CLIProxyManager.swift
@@ -17,6 +17,12 @@ final class CLIProxyManager {
     /// This solves the stale connection issue by forcing "Connection: close" on all requests
     let proxyBridge = ProxyBridge()
     
+    /// Closes the persistent traffic-statistics session whenever this manager
+    /// tears the proxy down, so CLIProxyAPI's in-memory counters restarting from
+    /// zero does not discard what the previous process served.
+    /// Driven from stop(), stopAndWait(), restartProxyIfRunning(), promote() and rollback().
+    @ObservationIgnored let usageSessionRecorder = ProxyUsageSessionRecorder()
+
     /// When enabled: clients connect to userPort, ProxyBridge forwards to internalPort
     /// When disabled: clients connect directly to userPort where CLIProxyAPI runs
     var useBridgeMode: Bool {
@@ -258,6 +264,12 @@ final class CLIProxyManager {
 
         Task {
             NSLog("[CLIProxyManager] Restarting proxy to apply configuration changes...")
+            // Fold a final usage snapshot before the process goes down.
+            // Reentrancy: this Task body runs after the guard above, so the
+            // running state is re-read here rather than reused from it.
+            if proxyStatus.running {
+                await usageSessionRecorder.proxyWillStop()
+            }
             stop()
             // Wait 0.5s for ports to clear
             try? await Task.sleep(nanoseconds: 500_000_000)
@@ -956,6 +968,11 @@ final class CLIProxyManager {
     }
     
     func stop() {
+        // Close the traffic-statistics session for the process being torn down.
+        // This path is synchronous, so no final counter sample can be fetched;
+        // callers that can await run proxyWillStop() first, which folds one.
+        usageSessionRecorder.proxyDidStop()
+
         terminateAuthProcess()
         stopHealthMonitor()
         cancelCrashRestart()
@@ -1009,6 +1026,13 @@ final class CLIProxyManager {
     /// Used by recovery paths that immediately restart, so the detached cleanup in stop()
     /// cannot race with and kill the newly started process by port.
     func stopAndWait() async {
+        // Fold a final usage snapshot and close the statistics session before
+        // the process is signalled. Deliberately the first statement: it
+        // suspends, so every piece of state captured below is read afterwards.
+        if proxyStatus.running {
+            await usageSessionRecorder.proxyWillStop()
+        }
+
         terminateAuthProcess()
         stopHealthMonitor()
         cancelCrashRestart()
@@ -1974,13 +1998,20 @@ extension CLIProxyManager {
             testConfigPath = nil
         }
         testManagementKey = nil
-        
+
+        // Fold a final usage snapshot before the old binary is torn down, so an
+        // upgrade does not reset the dashboard traffic statistics (issue #359).
+        // Reentrancy: this suspends, so `wasRunning` is read after it, not before.
+        if proxyStatus.running {
+            await usageSessionRecorder.proxyWillStop()
+        }
+
         // Stop the current active proxy if running
         let wasRunning = proxyStatus.running
         if wasRunning {
             stop()
         }
-        
+
         // Update the current symlink
         try storageManager.setCurrentVersion(version, source: source)
         activeVersion = version
@@ -2002,7 +2033,13 @@ extension CLIProxyManager {
         }
         
         managerState = .rollingBack
-        
+
+        // Fold a final usage snapshot before the process goes down.
+        // Reentrancy: this suspends, so `wasRunning` is read after it, not before.
+        if proxyStatus.running {
+            await usageSessionRecorder.proxyWillStop()
+        }
+
         // Stop current proxy
         let wasRunning = proxyStatus.running
         if wasRunning {
--- a/Quotio/Services/UsageStatsAggregator.swift
+++ b/Quotio/Services/UsageStatsAggregator.swift
@@ -1,0 +1,214 @@
+//
+//  UsageStatsAggregator.swift
+//  Quotio - Persistent Traffic Statistics
+//
+//  CLIProxyAPI keeps its /v0/management/usage counters in memory, so every
+//  proxy restart (or CLIProxyAPI upgrade) resets them to zero. This service
+//  accumulates those counters on the Quotio side and persists them to disk,
+//  so the dashboard traffic statistics survive proxy restarts and upgrades.
+//
+
+import Foundation
+
+// MARK: - Persisted Totals
+
+/// Fixed-size aggregate usage counters (no per-request entries, so the
+/// persisted file never grows with traffic).
+nonisolated struct UsageTotals: Codable, Sendable, Equatable {
+    var totalRequests: Int
+    var successCount: Int
+    var failureCount: Int
+    var totalTokens: Int
+    var inputTokens: Int
+    var outputTokens: Int
+
+    static let zero = UsageTotals(
+        totalRequests: 0,
+        successCount: 0,
+        failureCount: 0,
+        totalTokens: 0,
+        inputTokens: 0,
+        outputTokens: 0
+    )
+
+    init(
+        totalRequests: Int,
+        successCount: Int,
+        failureCount: Int,
+        totalTokens: Int,
+        inputTokens: Int,
+        outputTokens: Int
+    ) {
+        self.totalRequests = totalRequests
+        self.successCount = successCount
+        self.failureCount = failureCount
+        self.totalTokens = totalTokens
+        self.inputTokens = inputTokens
+        self.outputTokens = outputTokens
+    }
+
+    /// Snapshot the counters reported by CLIProxyAPI.
+    init(stats: UsageStats) {
+        let usage = stats.usage
+        self.init(
+            totalRequests: usage?.totalRequests ?? 0,
+            successCount: usage?.successCount ?? 0,
+            failureCount: usage?.failureCount ?? 0,
+            totalTokens: usage?.totalTokens ?? 0,
+            inputTokens: usage?.inputTokens ?? 0,
+            outputTokens: usage?.outputTokens ?? 0
+        )
+    }
+
+    static func + (lhs: UsageTotals, rhs: UsageTotals) -> UsageTotals {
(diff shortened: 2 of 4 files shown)


nguyenphutrong (changes requested):
Thank you for tackling this—keeping a fixed-size aggregate on disk is a good direction, and the synthetic merge with current `master` builds successfully with all 79 tests passing. I found two correctness blockers before this can reliably preserve traffic statistics:

1. `record(_:)` can still lose traffic across a restart. It infers a new proxy session only when the next observed `total_requests` is lower than the last polled value, and the proxy lifecycle never finalizes the current session. For example, if the last poll stored 100, nine more requests complete, the proxy restarts, and the first new sample is 1, the result is 101 rather than 110. If the last poll stored 5 and the restarted proxy reaches 10 before the next poll, the decrease is not observed at all, so the previous five requests are replaced. Please integrate with the managed stop/restart/upgrade lifecycle (including taking/folding a final snapshot where possible), or introduce another session boundary that makes the promised restart retention reliable rather than depending solely on a counter decrease.

2. The singleton store has no proxy/source identity, while `QuotaViewModel.refreshData()` is used for both local and remote proxy modes. Switching from a local snapshot of 100 requests to a remote endpoint at 20 folds the local 100 into the baseline and displays 120; switching back to the local endpoint at 100 then displays 200. Changing the configured remote server has the same problem, and the mixed total is persisted across launches. Please scope persisted state by the actual proxy endpoint/identity, or limit this aggregation to the intended local proxy and keep remote snapshots separate.

Please also add regression tests for these lifecycle and source-switch sequences. Once those cases are addressed, the persistence structure itself should be straightforward to reassess.

giaBaoJS:
Thanks — both blockers were real. Pushed `e10ddcd`; PR body updated with the full write-up.

**1. Restart retention no longer depends on a counter decrease.**

`CLIProxyManager` now owns a `ProxyUsageSessionRecorder` and drives it from every path on which it tears the managed proxy down:

| Lifecycle path | `CLIProxyManager.swift` | Hook |
| --- | --- | --- |
| `stop()` — user toggle, mode switch (`QuotaViewModel.stopProxy()` :1448, `SettingsScreen` :1500) | `stop()` :970 | `proxyDidStop()` (synchronous, no fetch possible) |
| `stopAndWait()` — health recovery (`QuotaViewModel.attemptProxyRecovery()` :1524) | `stopAndWait()` :1028 | `await proxyWillStop()` |
| Config restart — port, host, allow-remote, logging, routing strategy, proxy URL | `restartProxyIfRunning()` :262 | `await proxyWillStop()` → `stop()` |
| **Upgrade** — `performManagedUpgrade(to:)` :1795 → promote | `promote(version:source:)` :1987 | `await proxyWillStop()` → `stop()` |
| Rollback | `rollback()` :2030 | `await proxyWillStop()` → `stop()` |

`proxyWillStop()` fetches a final `/v0/management/usage` sample from the process about to go down (closure installed by `QuotaViewModel`, which owns the management client), folds it, then closes the session — baseline absorbs the session, live sample is zeroed. The next process can start from zero, or from anything, without discarding what the previous one served. `proxyDidStop()` closes without a fetch and is idempotent, so paths running both fold exactly once. The counter-decrease check is kept, but only as a fallback for restarts Quotio does not perform.

Your two numeric cases are now tests:

- `testManagedRestartFoldsFinalSnapshotTakenAtTeardown` — 100 stored, +9 unpolled, managed restart, first new sample 1 → **110**
- `testManagedRestartRetainsTotalsWhenNoDecreaseIsEverObserved` — 5 stored, restarted proxy already at 10 by the next poll, decrease never observed → **15**

**2. Persisted state is now scoped by proxy identity.**

New `UsageStatsSource`: `.localProxy`, and `.remote(endpoint:)` keyed on the normalized management base URL. Separate buckets that are never summed. `refreshData()` resolves the source *together with* the `apiClient` it belongs to (before the awaits), and `QuotaViewModel.initialize()` — which every mode switch routes through — re-scopes the displayed totals. Version 1 files migrate onto the local bucket.

- `testLocalAndRemoteTotalsAreNeverMerged` — local 100 → remote 20 shows **20**, not 120; back to local 100 shows **100**, not 200
- `testDistinctRemoteServersKeepSeparateTotals`, `testMixedSourceTotalsStaySeparateAcrossRelaunch`, `testLifecycleBoundaryOnlyClosesLocalProxySession`

**Limitation, stated plainly.** Retention is exact for restarts Quotio performs. It is not exact otherwise, and the PR no longer implies it is:

- Managed stop / config restart / upgrade / rollback / health recovery — exact; the final snapshot covers requests served since the last poll.
- Synchronous `stop()` — the session closes correctly, but no final snapshot is taken, so up to one refresh interval of requests goes uncounted.
- Proxy dies without Quotio's involvement — crash, external kill, sleep/power loss, or `terminateProxyOnShutdown()` at app quit. There is no opportunity to snapshot, so traffic served since the last poll is genuinely unrecoverable. The decrease fallback preserves everything up to that poll and nothing after.

**Concurrency.** All `@MainActor`. `proxyWillStop()` re-reads the stored sample *after* the fetch suspension and merges component-wise rather than overwriting, so a refresh landing mid-fetch is neither lost nor double-counted (`testConcurrentPollDuringFinalSnapshotIsMergedNotDoubleCounted`). In `stopAndWait()`, `promote()`, `rollback()` and `restartProxyIfRunning()` the hook is awaited *before* `proxyStatus.running` and process state are captured, never after.

**One thing I did not do:** there is no test that calls `CLIProxyManager.stop()` itself. `stop()` runs `killProcessOnPortSync(proxyStatus.port)` (:1251), which `lsof`s the configured port and kills whatever holds it — running that in the test suite would kill an unrelated process on the developer's or CI machine. So the hook is tested directly against `ProxyUsageSessionRecorder`, the same object the manager drives, and the call sites are listed above for review. Happy to change the approach if you'd prefer it covered differently.

Build and tests: `xcodebuild ... build` **BUILD SUCCEEDED**; `xcodebuild test ... -destination 'platform=macOS'` **TEST SUCCEEDED**, 84 passed / 0 failed.

nguyenphutrong (changes requested):
The original restart-retention and source-scoping cases are substantially addressed, but one lifecycle race remains. ProxyUsageSessionRecorder.proxyWillStop() suspends without marking the session as closing, so two restart/teardown tasks can fetch the same process concurrently. The first folds and closes the session; the second then folds the same-session sample into the zeroed lastSample, doubling the persisted baseline. This is reachable because multiple configuration setters launch independent restart tasks while proxyStatus.running remains true during the fetch. Please serialize finalization or use a session generation/token, and add an overlapping proxyWillStop regression test. The synthetic merge builds and the current full suite passes.

nguyenphutrong:
Thank you for the contribution. This branch now conflicts with master, and the lifecycle-based traffic statistics persistence proposed here is no longer planned in this form. Resolving the conflicts would not be enough because the implementation direction would also need to change, so this PR is being closed. The linked issue will remain available for a future approach.

PR #481, opened 2026-08-12
feat(providers): custom HTTP headers for OpenAI/Claude/Codex-compatible custom providers

### LINKED ISSUE #212: [Feature] Add Support for Custom Header Fields for Privately Deployed LLMs
### Problem Statement

When integrating with privately deployed large language models (LLMs), many deployments use custom HTTP header fields for authentication, routing, or version control (e.g., X-API-Key, X-Model-Version, X-Tenant-ID). Currently, the system only supports a standard set of headers, forcing users to modify the core HTTP client or use workarounds, which is inefficient and error-prone

### Proposed Solution

Extend the HTTP client configuration to allow users to define custom headers key-value pairs. This could be implemented via:

1. A configuration file (e.g., config.yml) section for custom headers.
2. Environment variable support (e.g., CUSTOM_HEADERS='{"X-Model-Version":"2024-12"}').
3. Programmatic API override for advanced use cases.
The goal is to maintain backward compatibility while providing flexibility for private deployments

### Alternative Solutions

_No response_

### Feature Category

Settings / Configuration

### Use Case

This feature is essential for enterprise and research environments where LLMs are hosted internally. Common scenarios include:

Authentication: Private endpoints often use custom headers like X-API-Key instead of standard Authorization.
Multi-tenancy: Headers like X-Tenant-ID route requests to specific model instances.
A/B Testing: Headers like X-Model-Variant can specify which model version to query.
Without this feature, users must fork the project or maintain fragile patches, increasing maintenance burden and risk of drift from upstream.

### Mockups / Examples

_No response_

### Additional Context

- This is a common pattern seen in tools like private OpenAI-compatible servers, vLLM, and TGI deployments.

- Implementing this would make the project more versatile and attractive for enterprise adoption.
- I am willing to contribute a pull request if the feature request is accepted and guidance is provided on the preferred implementation approach

### Checklist

giaBaoJS:
## Motivation

Private LLM deployments (vLLM, TGI, internal OpenAI-compatible gateways) frequently rely on custom HTTP headers for authentication, tenancy routing, or model versioning (`X-API-Key`, `X-Tenant-ID`, `X-Model-Version`, ...). Quotio already ships a custom header editor, but it is only enabled for **Gemini-compatible** providers — the exact provider types used for private deployments (OpenAI-compatible, Claude-compatible, Codex-compatible) could not configure headers at all.

Closes #212

## Proxy config format reference

CLIProxyAPI natively supports a `headers` map for all four provider types this PR enables, so the emitted config is honored end-to-end:

- `openai-compatibility` (provider level): [`OpenAICompatibility.Headers`](https://github.com/router-for-me/CLIProxyAPI/blob/2e6b1d83f6c304a102aa33c1faf0a4f94d0d331e/internal/config/config_types.go#L612-L613)
- `claude-api-key` (per key entry): [`ClaudeKey.Headers`](https://github.com/router-for-me/CLIProxyAPI/blob/2e6b1d83f6c304a102aa33c1faf0a4f94d0d331e/internal/config/config_types.go#L346-L347)
- `codex-api-key` (per key entry): [`CodexKey.Headers`](https://github.com/router-for-me/CLIProxyAPI/blob/2e6b1d83f6c304a102aa33c1faf0a4f94d0d331e/internal/config/config_types.go#L444-L445)
- `gemini-api-key` (per key entry, already supported by Quotio): [`GeminiKey.Headers`](https://github.com/router-for-me/CLIProxyAPI/blob/2e6b1d83f6c304a102aa33c1faf0a4f94d0d331e/internal/config/config_types.go#L533-L534)

Example emitted section:

```yaml
openai-compatibility:
  - name: "Private vLLM"
    base-url: "https://llm.internal/v1"
    headers:
      "X-API-Key": "secret"
      "X-Tenant-ID": "team-a"
    api-key-entries:
      - api-key: "sk-..."
```

The proxy applies these headers on upstream requests and documents that custom headers intentionally override built-in defaults ([`internal/util/header_helpers.go`](https://github.com/router-for-me/CLIProxyAPI/blob/main/internal/util/header_helpers.go)), which is exactly the issue's use case (e.g. `X-API-Key` instead of standard `Authorization`). Overriding auth headers is therefore allowed rather than blocked.

**Deliberately excluded:** `glm-api-key` (the section only exists in the bundled Plus binary and has no publicly documented `headers` support — emitting an unknown field there would be a guess) and ClinePass (managed service with fixed authentication).

## UX

The existing "Custom Headers" key/value rows in the Add/Edit Custom Provider sheet now appear for OpenAI-, Claude-, and Codex-compatible providers as well (same section, same styling, existing localized strings). Custom headers are also applied to the sheet's "Fetch models" and connection-test requests, so a private endpoint that requires them can be tested before saving.

Validation (localized in en/fr/vi/zh-Hans):
- header names must be RFC 7230 tokens (ASCII letters, digits, ``!#$%&'*+-.^_`|~``)
- duplicate names (case-insensitive) are rejected — the YAML `headers` map cannot express duplicates

Header values are stored exactly like API keys are today (same provider persistence), never logged, and YAML emission quotes keys and escapes `"`/`\` in values so arbitrary secret values cannot break the config.

## Implementation notes

- `CustomProviderType.supportsCustomHeaders` now returns true for openai/claude/gemini/codex compatibility types; the sheet and persistence key off it
- YAML emission places `headers` at the provider level for `openai-compatibility` and per key entry for `claude-api-key` / `gemini-api-key` / `codex-api-key`, matching the proxy structs above
- Invalid or empty header names are never emitted (defense in depth on top of validation)
- `CustomProviderService.removeCustomProviderSections(from:)` visibility changed from `private` to internal so tests can verify the strip/re-append round-trip

## Test evidence

New `QuotioTests/CustomProviderHeaderTests.swift` (17 tests): per-type YAML emission and placement, quoting/escaping, empty/invalid name filtering, RFC 7230 name validation, case-insensitive duplicate detection, Codable round-trip, and config strip/re-append idempotence (headers survive `syncToConfigFile`-style rewrites without corrupting the config).

```
** BUILD SUCCEEDED **

** TEST SUCCEEDED **   (82 tests passed, 0 failed)
```

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -4200,6 +4200,64 @@
         }
       }
     },
+    "customProviders.error.headerNameDuplicate" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Duplicate custom header names are not allowed"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Les noms d'en-têtes personnalisés en double ne sont pas autorisés"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Không được phép trùng tên header tùy chỉnh"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "自定义标头名称不能重复"
+          }
+        }
+      }
+    },
+    "customProviders.error.headerNameInvalid" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Custom header names may only contain letters, digits, and the characters !#$%&'*+-.^_`|~"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Les noms d'en-têtes personnalisés ne peuvent contenir que des lettres, des chiffres et les caractères !#$%&'*+-.^_`|~"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Tên header tùy chỉnh chỉ được chứa chữ cái, chữ số và các ký tự !#$%&'*+-.^_`|~"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "自定义标头名称只能包含字母、数字以及字符 !#$%&'*+-.^_`|~"
+          }
+        }
+      }
+    },
     "customProviders.fetchModels" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -4258,6 +4258,35 @@
         }
       }
     },
+    "customProviders.error.headerValueInvalid" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Custom header values must be printable ASCII, without line breaks or control characters"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Les valeurs d'en-têtes personnalisés doivent être en ASCII imprimable, sans saut de ligne ni caractère de contrôle"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Giá trị header tùy chỉnh phải là ký tự ASCII in được, không có ký tự xuống dòng hoặc ký tự điều khiển"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "自定义标头值必须为可打印的 ASCII 字符，不能包含换行符或控制字符"
+          }
+        }
+      }
+    },
     "customProviders.fetchModels" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/Models/CustomProviderModels.swift
+++ b/Quotio/Models/CustomProviderModels.swift
@@ -152,12 +152,16 @@ enum CustomProviderType: String, CaseIterable, Codable, Identifiable, Sendable {
         }
     }
 
-    /// Whether this provider type supports custom headers
+    /// Whether this provider type supports custom headers.
+    /// CLIProxyAPI accepts a `headers` map for `openai-compatibility` providers (provider level)
+    /// and for `claude-api-key`, `gemini-api-key`, and `codex-api-key` entries (per key).
+    /// GLM and ClinePass are excluded: `glm-api-key` has no documented headers support,
+    /// and ClinePass is a managed service with fixed authentication.
     var supportsCustomHeaders: Bool {
         switch self {
-        case .geminiCompatibility:
+        case .openaiCompatibility, .claudeCompatibility, .geminiCompatibility, .codexCompatibility:
             return true
-        case .openaiCompatibility, .claudeCompatibility, .codexCompatibility, .glmCompatibility, .clinePass:
+        case .glmCompatibility, .clinePass:
             return false
         }
     }
@@ -223,20 +227,30 @@ struct ModelMapping: Codable, Identifiable, Hashable, Sendable {
 
 // MARK: - Custom Header
 
-/// A custom HTTP header for Gemini-compatible providers
+/// A custom HTTP header sent with requests to the upstream provider
 struct CustomHeader: Codable, Identifiable, Hashable, Sendable {
     let id: UUID
     var key: String
     var value: String
-    
+
     init(id: UUID = UUID(), key: String, value: String) {
         self.id = id
         self.key = key
         self.value = value
     }
-    
+
     enum CodingKeys: String, CodingKey {
         case id, key, value
+    }
+
+    /// Non-alphanumeric characters allowed in an HTTP header field name (RFC 7230 token)
+    private static let validNameSpecials = Set("!#$%&'*+-.^_`|~")
+
+    /// Whether the given string is a valid HTTP header field name (RFC 7230 token: ASCII letters, digits, and select punctuation)
+    static func isValidName(_ name: String) -> Bool {
+        !name.isEmpty && name.allSatisfy { character in
+            character.isASCII && (character.isLetter || character.isNumber || validNameSpecials.contains(character))
+        }
     }
 }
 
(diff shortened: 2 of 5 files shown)


nguyenphutrong (changes requested):
Thank you—upstream placement is correct for all four supported provider families, and the synthetic merge with current `master` builds successfully with all 87 tests passing. I found two input-normalization blockers before this is safe for arbitrary private-endpoint headers:

1. Header names are validated and emitted after trimming, but the raw names are retained and passed to both `URLRequest` paths (`CustomProviderSheet.swift` around lines 865–868 and 1024–1027). Thus a user-entered name such as ` X-Tenant-ID ` is accepted by validation and will later generate the correct YAML key, yet the mandatory pre-save connection test sends the whitespace-bearing name. I reproduced this with CFNetwork: it serialized the line as a continuation of `Host`, so the server observed a corrupted Host value and no `X-Tenant-ID` header. The Fetch Models path also applies form headers before running the new invalid/duplicate-name validation. Please normalize once before persistence and reuse the same validated header representation for YAML, model fetch, and connection testing (or reject surrounding whitespace consistently).

2. `escapedYAMLString(_:)` escapes only backslashes and quotes. Literal CR/LF or other control characters remain in a quoted YAML scalar; a newline value such as `line1\nline2` is parsed as `line1 line2`, changing the secret, while other controls can make the generated CLIProxyAPI config invalid. CFNetwork may also silently omit a header with CR/LF, and Go's outbound HTTP stack rejects such values. Please validate values against wire-safe HTTP header-value rules and perform complete YAML double-quoted escaping for all accepted characters. Add parser round-trip tests for whitespace/control cases rather than only asserting generated substrings.

These should be addressed together so the editor, its direct requests, persisted data, and generated proxy config all agree on the exact accepted header set.

giaBaoJS:
Both blockers addressed in 00d342a. Thanks for the CFNetwork repro — it pointed straight at the real defect, which was that four consumers each re-derived the header list independently.

**Blocker 1 — normalize once, reuse everywhere**

There is now exactly one canonical representation and every consumer reads it:

- `CustomHeader.canonicalized(_:)` (`Quotio/Models/CustomProviderModels.swift:291`) strips RFC 7230 optional whitespace and drops unnamed entries. It trims **space and tab only**, not `CharacterSet.whitespaces`, so a value containing U+00A0 or a newline is not silently rewritten into something that passes — it is rejected instead.
- `CustomProvider.effectiveHeaders` (`CustomProviderModels.swift:486`) is the single accessor used for output. It gates on `supportsCustomHeaders`, canonicalizes, then filters to wire-safe name/value pairs. That filter is defence in depth: `validate()` refuses to save anything it would drop, but a provider decoded from storage that predates this validation must still not build a corrupt request or an invalid config.
- `URLRequest.applyCustomHeaders(_:)` (`CustomProviderModels.swift:331`) is the single request-mutation point, so the two `URLRequest` paths cannot drift again.
- Fetch Models now canonicalizes and runs `CustomHeader.validationErrors(in:)` **before** the request is built, returning the error into `modelFetchError` instead of sending anything.

Consumer inventory (every read of `CustomProvider.headers` / the sheet's `@State headers`):

| Consumer | File:line | Now uses |
|---|---|---|
| Validation | `CustomProviderModels.swift:465` (`validate()`) | `CustomHeader.validationErrors(in: CustomHeader.canonicalized(headers))` — the unfiltered canonical list, so it can *report* what the emitter would drop |
| YAML emission (all 4 families) | `CustomProviderModels.swift:516` (`headersYAML(indent:)`, called at :579, :617, :649, :669) | `effectiveHeaders` |
| Pre-save connection test | `CustomProviderSheet.swift:1032` (`testConnection`) | `request.applyCustomHeaders(provider.effectiveHeaders)` |
| Fetch Models | `CustomProviderSheet.swift:875` (`fetchModelsFromAPI`) | `request.applyCustomHeaders(customHeaders)`, canonicalized at :846 and validated at :848 before the request is built |
| Persistence | `CustomProviderSheet.swift:948` (`saveProvider`) | `CustomHeader.canonicalized(headers)` |
| Storage round-trips | `CustomProviderService.swift:44/70/99` (add/update/toggle) | pass through the already-canonical array unchanged |

`GLMAPIKeySheet.swift:210` also constructs a `CustomProvider`, but with `glmCompatibility` and no headers; `effectiveHeaders` returns `[]` for it and for ClinePass regardless of stored contents (test: `testEffectiveHeadersAreEmptyForUnsupportedProviderTypes`).

**Blocker 2 — value validation and complete YAML escaping**

- New `CustomHeader.isValidValue(_:)` (`CustomProviderModels.swift:270`) enforces RFC 7230 `field-value`: HTAB, SP, and visible US-ASCII. CR/LF, other C0/C1 controls, DEL, and non-ASCII are rejected. This also closes header injection through a *value* (`"team-a\r\nAuthorization: Bearer stolen"`), which the name-only token check did not cover.
- `escapedYAMLString(_:)` (`CustomProviderModels.swift:536`) now implements the full YAML double-quoted escape set (`\0 \a \b \t \n \v \f \r \e \N \_ \L \P`, plus `\xNN` for remaining C0/C1/DEL, alongside `\\` and `\"`), so emission is lossless for any string it is handed even though validation rejects those characters upstream.
- New localized string `customProviders.error.headerValueInvalid` for en / fr / vi / zh-Hans, inserted textually into `Localizable.xcstrings`.

I did **not** touch `escapedName` (`CustomProviderModels.swift:698`) or the other interpolated fields (`base-url`, `prefix`, `api-key`), which have the same partial-escaping shape. They predate this PR and fixing them is a separate change — happy to open a follow-up if you want it.

**Tests** — `QuotioTests/CustomProviderHeaderTests.swift`, 30 cases (was 17). Per your note, these parse rather than substring-match: `YAMLHeaderBlockReader` decodes double-quoted scalars using the YAML 1.2 escape table.

- `testConnectionTestRequestSendsNormalizedHeaderName` — your exact repro: `" X-Tenant-ID "` / `" team-a "` reaches `URLRequest` as `X-Tenant-ID: team-a`, and asserts every name actually on the request is an RFC 7230 token.
- `testRequestHeadersMatchGeneratedYAMLHeaders` — the Fetch Models request, the connection-test request, and the parsed YAML all carry identical name/value pairs.
- `testGeneratedYAMLRoundTripsBackToTheCanonicalHeaders` — quotes, backslashes, embedded tabs, and `a: b # not-a-comment` parse back to the exact `effectiveHeaders`.
- `testYAMLEscapingRoundTripsControlAndWhitespaceCharacters` — `\n`, `\r\n`, NUL, DEL, ESC, U+0085, U+00A0, U+2028 all survive escape → parse unchanged and never emit a raw line break.
- `testCanonicalizedDoesNotStripNewlinesFromValues`, `testValidateRejectsCRLFHeaderInjectionAttempt`, `testValidateRejectsHeaderValueWithLineBreak`, `testNormalizationMakesWhitespacePaddedDuplicatesDetectable`, `testEffectiveHeadersDropEntriesThatValidationWouldReject`.

**Verification** — `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` succeeds; `xcodebuild test -destination 'platform=macOS'` is green with 95 tests executed.

nguyenphutrong (approved):
The security/data-integrity blockers are resolved. Header normalization is centralized across validation, persistence, both request paths, and YAML generation; unsafe names/values are rejected or filtered, CR/LF injection is closed, and YAML escaping round-trips the accepted representation. The synthetic merge and targeted CustomProviderHeaderTests pass.

PR #482, opened 2026-08-12
feat(opencode): split Gemini models into a Google-protocol provider

giaBaoJS:
## Motivation

From #227:

> Right now, autoconfiguring Opencode will put all models in a Anthropic-flavor provider. This works great for Claude, but gives Gemini models on Antigravity an identity crisis where they start spending time on contemplating whether they are Opencode or Antigravity instead of doing the task.
>
> This is reported not to happen when using those models on Google's protocol on the /v1beta endpoint. Make it so that the automatic config for Opencode creates 2 separate providers for the assorted models.

`generateOpenCodeConfig` previously emitted a single `provider.quotio` entry with `npm: "@ai-sdk/anthropic"` and `baseURL: <proxy>/v1`, and dumped every model into it — including the Gemini-native ones (`gemini-3-pro-preview`, `gemini-2.5-flash`, ...) that Antigravity exposes alongside its `gemini-claude-*` Claude models.

## Protocol / endpoint reference

CLIProxyAPI exposes a Gemini-native (Google protocol) surface on `/v1beta` in addition to the OpenAI/Anthropic-compatible `/v1` routes — `internal/api/server_routes.go`:

```go
// Gemini compatible API routes
v1beta := s.engine.Group("/v1beta")
v1beta.Use(AuthMiddleware(s.accessManager))
{
    v1beta.GET("/models", s.geminiModelsHandler(geminiHandlers))
    v1beta.POST("/interactions", geminiHandlers.Interactions)
    v1beta.POST("/models/*action", geminiHandlers.GeminiHandler)
    v1beta.GET("/models/*action", s.geminiGetHandler(geminiHandlers))
}
```

So the endpoint the reporter describes really is available locally, and OpenCode can reach it through the `@ai-sdk/google` provider package.

## What changed

When OpenCode config is generated, models are partitioned deterministically by id:

- **`provider.quotio`** — unchanged: `@ai-sdk/anthropic`, `<proxy>/v1`, `litellmProxy: true`. Holds Claude models (including Antigravity's `gemini-claude-*` ids) and everything else (GPT, Qwen, MiniMax, ...).
- **`provider.quotio-gemini`** — new: `@ai-sdk/google`, `<proxy>/v1beta`. Holds Gemini-native models only.

The Gemini entry is emitted **only when Gemini-native models are present**, so non-Antigravity setups produce exactly the same single-provider config as today.

[...]

PR #483, opened 2026-08-12
feat(dashboard): show available models in the app panel

### LINKED ISSUE #201: The should show what models can be used in the panel
### Problem Statement

The should show what models can be used in the panel, istead of curling myself

### Proposed Solution

show all the models I can use in the panel

### Alternative Solutions

_No response_

### Feature Category

New AI Provider Integration

### Use Case

_No response_

### Mockups / Examples

_No response_

### Additional Context

_No response_

### Checklist

giaBaoJS:
## Motivation

From #201:

> The should show what models can be used in the panel, istead of curling myself
>
> show all the models I can use in the panel

Today the proxy's model list is only reachable inside modal flows — the agent config sheet, the Fallback "add entry" sheet, and the custom provider sheet all call `/v1/models`, but none of them is a place you go to *read* the list. To find out which model IDs are routable, users end up curling the endpoint themselves.

This adds a first-class **Available Models** section that lists them.

## What this adds

- **Dashboard → Available Models**: model IDs grouped by provider, fetched live from the proxy's OpenAI-compatible `/v1/models` endpoint, with a refresh button in the section header and a model count.
- **Click to copy**: model IDs exist to be pasted into client configs, so each row copies its ID to the pasteboard and shows a short "Copied" confirmation.
- **Explained empty states**: a note when the proxy is stopped ("Start the proxy to see the models you can use."), when the proxy reports no models, and when the fetch fails.
- **Localized** in en / fr / vi / zh-Hans, using the existing `Localizable.xcstrings` approach.

## Placement rationale

The section sits on the Dashboard, directly below the existing **API Endpoint** row in `fullModeContent`. That pairing is the point: the endpoint tells you *where* to send requests, the model list tells you *what* to ask for — together they are everything you need to paste into a client config. It reuses the existing `GroupBox` idiom of the neighbouring sections, so no navigation changes, no new tab, and no `NavigationPage` case were needed.

The section is scoped to local-proxy mode, where a `/v1/models` endpoint actually exists; Monitor and remote modes are untouched.

## Implementation notes

- Model fetching reuses `AgentSetupViewModel.loadModels(forceRefresh:)`, so the Dashboard shares the same cached list as the Fallback and agent-config flows rather than opening a second code path to the proxy. No networking was added to the view.
- The response parsing and provider grouping were extracted into a new pure `ModelCatalog` helper (`Quotio/Models/ModelCatalog.swift`). `AgentConfigurationService.fetchAvailableModels` now calls `ModelCatalog.parse` instead of carrying its own nested `Decodable` struct — same behaviour, including the GitHub Copilot availability filter and the `owned_by` → `"openai"` default, now covered by tests.
- Grouping is derived from the proxy's own `owned_by` field, which is the authoritative answer for "which provider serves this model", rather than inferring it from the model ID.

## Test evidence

New `QuotioTests/ModelCatalogTests.swift` covers response parsing (well-formed, missing `owned_by`, empty list, malformed payload), provider grouping and sorting, duplicate-ID collapsing, the same ID served by two providers, the empty case, and a parse → group end-to-end case.

```
** BUILD SUCCEEDED **

** TEST SUCCEEDED **

Test suite 'ModelCatalogTests' started
Test case 'ModelCatalogTests.testGroupByProviderCollapsesDuplicateModelIdsWithinProvider()' passed
Test case 'ModelCatalogTests.testGroupByProviderGroupsAndSortsModels()' passed
Test case 'ModelCatalogTests.testGroupByProviderKeepsSameModelIdOfferedByDifferentProviders()' passed
Test case 'ModelCatalogTests.testGroupByProviderReturnsEmptyForNoModels()' passed
Test case 'ModelCatalogTests.testParseDefaultsMissingOwnedByToOpenAI()' passed
Test case 'ModelCatalogTests.testParseReadsIdAndOwnedByFromModelsResponse()' passed
Test case 'ModelCatalogTests.testParseReturnsEmptyListForEmptyResponse()' passed
Test case 'ModelCatalogTests.testParseThenGroupProducesProviderSectionsEndToEnd()' passed
Test case 'ModelCatalogTests.testParseThrowsOnMalformedResponse()' passed
```

The existing `AmpQuotaFetcherTests` and `MonitorRuntimeTests` suites also pass on this branch.

Closes #201

### DIFF
--- a/Quotio/Models/ModelCatalog.swift
+++ b/Quotio/Models/ModelCatalog.swift
@@ -3,26 +3,135 @@
 //  Quotio
 //
 //  Pure helpers for parsing the proxy's OpenAI-compatible /v1/models
-//  response and grouping the resulting models by provider for display.
+//  response, plus the display state that keeps a live catalog readout
+//  distinguishable from a stale one.
 //
 
 import Foundation
 
-/// Models exposed by the proxy for a single provider.
-nonisolated struct ProviderModelGroup: Identifiable, Equatable, Sendable {
-    let provider: String
-    let models: [AvailableModel]
+/// One entry of the proxy's OpenAI-compatible `/v1/models` response.
+///
+/// This is deliberately a thin mirror of the wire format. It carries only what
+/// `/v1/models` actually states: the model ID, and the `owned_by` metadata when
+/// present. It says nothing about which provider account will serve a request
+/// for the ID — the endpoint does not expose that.
+nonisolated struct ModelCatalogEntry: Identifiable, Equatable, Sendable {
+    /// The model ID exactly as listed by `/v1/models`.
+    let id: String
 
-    var id: String { provider }
+    /// Raw `owned_by` value, or `nil` when the response omitted the field.
+    ///
+    /// This is static model-owner metadata. CLIProxyAPI deduplicates `/v1/models`
+    /// by ID and tracks serving providers separately, so the value here may be
+    /// replaced when another provider registers the same ID. Do not treat it as
+    /// routing provenance.
+    let owner: String?
+
+    /// Owner metadata fit for display, or `nil` when the endpoint reported
+    /// nothing usable (field absent, empty, or whitespace only).
+    var displayOwner: String? {
+        guard let trimmed = owner?.trimmingCharacters(in: .whitespacesAndNewlines),
+              !trimmed.isEmpty
+        else { return nil }
+        return trimmed
+    }
+}
+
+/// How the catalog currently on screen relates to the proxy's live state.
+nonisolated enum ModelCatalogFreshness: Equatable, Sendable {
+    /// Nothing has been fetched successfully yet.
+    case never
+
+    /// The entries are exactly what `/v1/models` returned at `fetchedAt`.
+    case live(fetchedAt: Date)
+
+    /// A previous successful response is still on screen because a later
+    /// fetch failed; it may no longer match the proxy.
+    case stale(fetchedAt: Date)
+}
+
+/// Display state for the read-only model catalog.
+///
+/// Unlike `AgentSetupViewModel.availableModels`, this never substitutes
+/// `AvailableModel.allModels` (or any other built-in list) for a missing or
+/// failed response: a successful empty response stays empty, and a failed
+/// refresh is reported as such instead of being papered over. That distinction
+/// is what lets the UI claim "live" only when it really is.
+nonisolated struct ModelCatalogState: Equatable, Sendable {
+    /// Entries currently on screen, deduplicated by ID and sorted.
+    private(set) var entries: [ModelCatalogEntry] = []
+
+    private(set) var freshness: ModelCatalogFreshness = .never
+
+    /// A fetch is in flight.
+    private(set) var isLoading = false
+
+    /// The most recent fetch attempt failed.
+    private(set) var lastFetchFailed = false
+
+    /// At least one fetch attempt has completed (successfully or not).
+    ///
+    /// Distinguishes "not fetched yet" from "fetched, and the proxy listed
+    /// nothing" so the empty state is reachable.
+    private(set) var hasCompletedFetch = false
+
+    /// The proxy answered successfully and listed no models at all.
+    var isEmptyLiveResult: Bool {
+        hasCompletedFetch && !lastFetchFailed && entries.isEmpty
+    }
+
+    mutating func beginLoading() {
+        isLoading = true
+    }
+
+    /// Records a successful `/v1/models` response verbatim.
+    mutating func apply(entries: [ModelCatalogEntry], fetchedAt: Date) {
+        self.entries = ModelCatalog.displayEntries(entries)
+        freshness = .live(fetchedAt: fetchedAt)
+        isLoading = false
+        lastFetchFailed = false
+        hasCompletedFetch = true
+    }
+
+    /// Records a failed fetch, keeping any earlier response but demoting it to stale.
+    mutating func applyFailure() {
+        isLoading = false
+        lastFetchFailed = true
+        hasCompletedFetch = true
+
+        switch freshness {
+        case .never:
+            break
+        case .live(let fetchedAt), .stale(let fetchedAt):
+            freshness = .stale(fetchedAt: fetchedAt)
+        }
+    }
+
+    /// Drops everything, e.g. when the proxy stops and nothing on screen can be trusted.
+    mutating func reset() {
+        self = ModelCatalogState()
+    }
+}
+
+nonisolated enum ModelCatalogError: LocalizedError, Equatable {
+    /// No running proxy to ask, so no catalog can be fetched.
+    case proxyUnavailable
+
+    var errorDescription: String? {
+        switch self {
+        case .proxyUnavailable:
+            return "The proxy is not available."
+        }
+    }
 }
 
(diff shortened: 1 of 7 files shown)


Quotio/Views/Components/AvailableModelsSection.swift
```diff
@@ -0,0 +1,171 @@
...
+        guard viewModel.agentSetupViewModel.availableModels.isEmpty else { return }
+        await loadModels(force: false)
+    }
+
+    private func loadModels(force: Bool) async {
+        guard isProxyRunning else { return }
+        isLoading = true
+        defer { isLoading = false }
+
+        let loadedFromRemote = await viewModel.agentSetupViewModel.loadModels(forceRefresh: force)
```
nguyenphutrong: `AgentSetupViewModel.loadModels` cannot currently provide the exact live catalog this view promises: `processModels` replaces a successful empty `/v1/models` response with `AvailableModel.allModels` and still returns `true`; after an earlier failed fetch it also leaves those defaults in `availableModels`, causing `loadModelsIfNeeded` to skip the remote request. In both cases this section can show models that are not routable, and the new empty state is unreachable. Please keep remote catalog data/status distinct from agent-setup fallbacks (or expose a fetch path that returns the exact response) and add end-to-end coverage for empty, failed, and stale/default-cache cases.

Quotio/Models/ModelCatalog.swift
```diff
@@ -0,0 +1,63 @@
...
+    let models: [AvailableModel]
+
+    var id: String { provider }
+}
+
+nonisolated enum ModelCatalog {
+
+    /// Decodes an OpenAI-compatible `/v1/models` response body into models.
+    ///
+    /// Each item's `owned_by` field becomes the model's provider,
```
nguyenphutrong: CLIProxyAPI's `owned_by` is static model-owner metadata, not the provider/account that can route the ID. The registry deduplicates `/v1/models` by ID, tracks serving providers separately, and may replace the exposed metadata when another provider registers the same ID. Therefore grouping this as the model's `provider` can be misleading, and the test that keeps one ID under two providers describes a response the endpoint cannot emit. Please present this as owner metadata (with empty-value handling) or omit the provider grouping unless the proxy exposes actual routing provenance.

nguyenphutrong (changes requested):
Thank you for turning the requested model list into a polished, localized Dashboard section. The PR merges cleanly into current `master`; the Debug build and all 80 tests pass. Changes are requested because the shared agent-setup fallback semantics can make this catalog present built-in or stale models as live, usable models, and because the provider grouping assigns routing meaning to metadata that `/v1/models` does not provide. Details are inline.

giaBaoJS:
Thanks — both points were right, and both are fixed in 5a7c29f.

**1. `AvailableModelsSection.swift:168` — agent-setup fallbacks leaking into the catalog**

Confirmed all three leaks you described in `AgentSetupViewModel.loadModels(forceRefresh:)`:
- `processModels` (`AgentSetupViewModel.swift:404-411`) returns `AvailableModel.allModels` for a successful empty `/v1/models` response, and `loadModels` still sets `loadedFromRemote = true` (`:386`) — so built-ins were shown and the section's empty state was unreachable.
- the `catch` at `:391-398` writes `AvailableModel.allModels` into `availableModels`, so the old `loadModelsIfNeeded` (`availableModels.isEmpty` guard) then skipped the remote request entirely.
- a third one I hit while tracing: `refreshVirtualModels()` (`:413-430`) appends locally configured virtual models with `provider == "fallback"`, which never came from `/v1/models` at all.

The section no longer touches that path. Added `AgentConfigurationService.fetchModelCatalog(config:)` and `AgentSetupViewModel.fetchModelCatalog()`, which return the response verbatim — no fallback list, no cache read or write, no mutation of `availableModels` — and throw on failure so the caller can tell a failure from an empty catalog. `fetchAvailableModels(config:)` is now a thin wrapper over it, so the picker path is byte-for-byte what it was.

State is tracked in `ModelCatalogState`, which keeps entries, freshness and failure separate, and the section labels itself accordingly:
- success → "Live from the proxy · fetched at HH:MM:SS"
- success with zero models → the empty note (now reachable), still marked live
- first failure → the error note, nothing listed
- failure after an earlier success → the earlier list is kept but marked "Refresh failed — showing the list fetched at HH:MM:SS, which may be out of date" in orange, with the original fetch time
- proxy stopped → state is reset so no leftover list sits under the "start the proxy" note

`loadIfNeeded` now gates on `hasCompletedFetch` rather than "list is empty", so an empty live catalog is not re-fetched as if it had never been fetched.

**2. `ModelCatalog.swift:23` — `owned_by` is not routing provenance**

Agreed; the provider grouping is gone, along with `ProviderModelGroup` and `groupByProvider`. The proxy exposes no routing provenance on `/v1/models`, so the section no longer implies any. Models are one flat list sorted by ID and deduplicated by ID (matching the registry's own by-ID deduplication). `owned_by` is now shown per row as owner metadata via `ModelCatalogEntry.displayOwner`, which is `nil` when the field is absent, empty or whitespace-only — in that case no owner is rendered rather than a blank or invented one. `parse` keeps `owned_by` as an optional raw value instead of defaulting it, and a footnote states that owner is metadata reported by `/v1/models` and does not indicate which provider account will serve a request.

The test that kept one ID under two providers is replaced by `testDisplayEntriesCollapsesRepeatedIdsKeepingFirstOccurrence`, which asserts the response the endpoint can actually emit.

**Shared-loader consumers verified unaffected**

`loadModels(forceRefresh:)` behavior is unchanged; only its endpoint/API-key setup moved into a private `modelRequestConfiguration()` helper, returning `nil` under exactly the condition that previously returned `false` early. The consumers I checked are `AgentConfigSheet.swift:341` (refresh button), `FallbackScreen.swift:33/119`, `WarmupSheet.swift:340` and the in-VM call at `AgentSetupViewModel.swift:123` (`startConfiguration`) — all still get the fallback list, the cross-dismissal cache and the virtual fallback models. `fetchAvailableModels(config:)` keeps its `owned_by ?? "openai"` default and the `provider == "github-copilot"` Copilot availability filter, both covered by `testAgentSetupMappingPreservesHistoricalOpenAIProviderDefault` and `testAgentSetupMappingPreservesResponseOrderAndDuplicates`.

**Coverage and verification**

`ModelCatalogTests` now covers parsing (absent/empty/whitespace `owned_by`, empty response, malformed body), display ordering and dedup, the full state machine (empty-live, first failure, success→failure→stale keeping the original timestamp, repeated failures, stale→live, refresh replacing rather than merging, reset), an assertion that no `AvailableModel.allModels` ID can appear in an empty live result, and the two agent-setup mapping guards above.

New strings (`availableModels.live`, `.stale`, `.owner`, `.ownerNote`) are localized in en, fr, vi and zh-Hans. `xcodebuild build -configuration Debug` and `xcodebuild test -destination 'platform=macOS'` both pass (89 tests, 0 failures, run twice; `MonitorRuntimeTests` green in both). Trial-merged clean against `master` and against #449, #467, #473, #477 and #482.

nguyenphutrong (approved):
The catalog now uses an independent live-fetch path without built-in, cached-picker, or virtual-model substitution. Empty, failed, live, and stale states are represented accurately, and owned_by is shown only as owner metadata rather than routing provenance. The synthetic merge and full suite pass.

Issue #128, opened 2026-01-07
closing Quotio does not clean/revert the Claude Code agent settings in settings.json


thesoulpole:
### What happened?

when you stop Quotio, the settings stay from Quotio. Więc Claude Code, when I want to use it NORMALLY witht the original subscription, 1) thinks its working in API mode, and 2) thinks the models it has are eg from Gemini, when in fact it has just the Anthropic models!

### Steps to reproduce

start, then stop Quotion and look what stays in .claude/settings.json

### Expected behavior

after stopping Quotio, the original settins should be reverted
either the original settings.json, or no such file if it did not exist before Quotio
IMPORTANT - when Quotio exits, it should save IT'S settings.json, and relaod it when it is restarted!

I do not know if same prblem is with OTHER Agents!

### Severity

🟡 Medium - Bug with workaround available

### Affected component

Agent Configuration

### Provider (if applicable)

None

### CLI Agent (if applicable)

Claude Code

### macOS Version

tahoe

### Quotio Version

0.6.0

### App Mode

Full Mode (Proxy running)

### Logs (optional)

```shell

```

miethe:
This issue does occur with other agents, with Codex being the most notable. I don't think there is going to be a particularly clean method for reverting changes on the fly, so I'd recommend for the app to back-up each config before updating it with the Quotio configs automatically, then swapping the config files whenever Proxy mode is turned on. Short of that, users should be instructed to back-up each of their local config files BEFORE configuring the agent in Quotio.

PR #484, opened 2026-08-12
feat(agents): restore agent configs on quit (opt-in)

giaBaoJS:
## Motivation

Addresses the revert-on-quit half of #128. It does **not** close the issue — see [Scope](#scope-relative-to-128) below.

When Quotio quits, the configs it wrote stay behind. `~/.claude/settings.json` still carries `ANTHROPIC_BASE_URL` pointing at the (now dead) proxy plus Quotio's model overrides, so Claude Code starts in API mode against nothing and offers model names it cannot serve. The same applies to the other agents Quotio configures — Codex is the one called out in the issue comments.

Quotio already knows how to undo this: every agent has a "Default" setup path that strips Quotio's keys and preserves the user's own settings, backing the file up first. It was just never run at quit time.

## Design

**Opt-in, default OFF.** A forced revert on every quit would be wrong for the equally common workflow of quitting and relaunching Quotio. This ships as a switch on the Agents screen — "Restore agent configs when Quotio quits" — off by default, which keeps today's behaviour byte-for-byte for everyone who does not turn it on.

**Reuses the existing revert paths.** `AgentConfigurationService.swift` is not modified. `AgentQuitRestoreService` orchestrates the quit-time work and calls `generateConfiguration(..., setupMode: .defaultSetup, mode: .automatic)` — the exact call the per-agent "Default" button makes. Key removal, TOML/JSON merging and timestamped backups are not duplicated, so per-agent fixes to that path (e.g. #467, #477, #482) apply here automatically once they land.

### Ownership: only revert what Quotio can prove it wrote

A `localhost` base URL says an agent talks to *some* local proxy, not that the proxy is Quotio's. Users run their own local gateways, so keying off `isProxyConfigured` would delete configuration Quotio never authored.

Ownership is now established from something Quotio itself authored, in `Quotio/Services/AgentConfigOwnership.swift`:

1. **Receipt** — when Quotio writes an agent config, it records the endpoint plus a **SHA-256 fingerprint** of the API key it issued (the raw key is never persisted). At quit time the on-disk endpoint *and* key are compared against that receipt. Match ⇒ these are Quotio's bytes. Receipt present but no longer matching ⇒ the user changed it by hand ⇒ **leave it alone** (`.externallyModified`).
2. **Quotio-authored marker** — fallback for configs written before receipts existed, and only where a marker is genuinely unambiguous.
3. Neither ⇒ **skip**. Not reverting is recoverable; deleting a config Quotio did not write is not.

Per-agent evidence, from what setup actually writes:

| Agent | Files setup writes | Quotio-authored in-file marker | Key evidence used by receipt |
|---|---|---|---|
| Claude Code | `~/.claude/settings.json` (`env.ANTHROPIC_*`, `model`); shell profile when `shellOnly`/`both` | **none** — `ANTHROPIC_*` are Anthropic's own variable names | `env.ANTHROPIC_AUTH_TOKEN` |
| Codex CLI | `~/.codex/config.toml`, `~/.codex/auth.json` | **yes** — `model_provider = "cliproxyapi"` + `[model_providers.cliproxyapi]` (`AgentConfigurationService.swift:340-355`) | `auth.json` → `OPENAI_API_KEY` |
| Amp CLI | `~/.config/amp/settings.json` (`amp.url`), `~/.local/share/amp/secrets.json` | **none** — `amp.url` is Amp's own key | `secrets.json` → `apiKey@<amp.url>` |
| OpenCode | `~/.config/opencode/opencode.json` | **yes** — `provider.quotio` / `"name": "Quotio"` (`AgentConfigurationService.swift:1198-1224`) | `provider.quotio.options.apiKey` |
| Factory Droid | `~/.factory/config.json` (`custom_models[]`) | **none** — entries carry no Quotio field | `custom_models[].api_key` |

So Codex and OpenCode can be recognised even without a receipt; **Claude Code, Amp and Factory Droid cannot**, and without a receipt they are skipped rather than guessed at. This is the case you raised directly: Factory Droid's default revert filters *every* localhost `custom_models` entry, so with no receipt Quotio now declines to run it at all rather than removing a user's own local model.

### Termination: the decision is deferred until the revert finishes

`applicationWillTerminate` cannot hold termination open, so the old three-second detached wait could let AppKit kill the process mid-loop. The sequence is now:

1. User quits → `applicationShouldTerminate(_:)`.
2. Toggle off → `.terminateNow`, nothing else happens. Second quit while a restore is already in flight → `.terminateNow` (the user is waiting on us).
3. Otherwise `AgentQuitRestoreTerminationCoordinator.begin()` starts two `@MainActor` tasks — the serial revert, and a **5 s watchdog** — and returns **`.terminateLater`**.
4. Whichever finishes first calls `resolve(_:)`, guarded by a one-shot flag on the main actor, then `NSApp.reply(toApplicationShouldTerminate: true)`. Completion logs the outcome; timeout logs the agents that *did* complete, from a lock-protected progress snapshot, so a wedge is diagnosable.
5. AppKit proceeds → `applicationWillTerminate` → proxy/tunnel teardown as before.

`.terminateLater` makes AppKit wait indefinitely, so a missed reply would hang the app and force a Force Quit — worse than the bug being fixed. The watchdog is what makes the deferral safe, and it is covered by a test that wedges an agent. Both tasks capture the coordinator strongly so a dropped reference cannot swallow the reply. 5 s is chosen to stay inside the shorter OS budget for a logout/restart-initiated quit.

Restoration now runs **before** `CLIProxyManager.terminateProxyOnShutdown()` rather than racing it.

**Interaction with #476:** `AppResetService.relaunch()` calls `exit(0)`, bypassing every termination handler including this one. That is the correct outcome — a reset relaunches Quotio immediately, so its agent configs should stay in place. No line-level overlap either; #476 does not touch `QuotioApp.swift`.

## Scope relative to #128

Narrowing this explicitly rather than over-claiming. This PR covers "stop the agent pointing at a dead proxy". It does **not** yet implement:

- **Restore the original file / delete it if Quotio created it.** The receipt records `primaryConfigExisted`, but the revert still strips known keys rather than restoring the prior file. Byte-exact restore stays available through the existing per-agent backup list.
- **Persist and reapply Quotio's own agent settings on the next launch.**
- **Auxiliary/alternate write targets**, which are gaps in the shared revert path, not in this orchestration:
  - Claude `shellOnly`/`both` writes exports to the shell profile. `ShellProfileManager.removeFromProfile` exists (`ShellProfileManager.swift:62`) but **has no callers anywhere in the repo** — the per-agent "Default" button does not undo shell exports either. With `shellOnly` no JSON is written, so this PR reads no config and skips the agent; it does not make anything worse, and it does not help.
  - Codex `auth.json` — being fixed in #467, whose hunks land in `generateCodexDefaultConfig`, the exact path this PR calls.
  - Amp `secrets.json` — `generateAmpDefaultConfig` removes `amp.url` but leaves the `apiKey@<url>` entry. Inert once the URL is gone, but still untidy.

Happy to fold any of these in here or split them out — say which you prefer.

## Test evidence

`QuotioTests/AgentConfigOwnershipTests.swift` (14 tests) — every test runs against a **throwaway temp home directory**; nothing reads or writes real agent configs:

- receipt proves ownership for all five agents, across the `/v1` and trailing-slash forms each file format stores
- a foreign localhost proxy (`http://localhost:4000`, user's own key) is **not owned** for Claude Code, Amp and Factory Droid
- Factory Droid: a foreign localhost entry alongside a Quotio receipt is not owned; a matching entry among several is
- rotated `ANTHROPIC_AUTH_TOKEN`, and a Codex `auth.json` key replaced outside Quotio ⇒ `.externallyModified`
- `cliproxyapi` / `provider.quotio` markers recognised without a receipt; foreign providers are not
- receipt round-trip, and the raw API key is absent from the persisted blob

`QuotioTests/AgentQuitRestoreTests.swift` (11 tests):

- unowned agents are never handed to the reverter (the decisive assertion)
- **all agents complete before the reply fires**
- **a wedged agent hits the timeout, still replies, and reports partial progress**
- **the reply fires exactly once** even when the revert finishes just after the deadline
- failure isolation for an unsuccessful result and for a thrown error
- `CLIAgent.allCases` coverage; setting round-trip and default-OFF against an isolated `UserDefaults` suite

```
xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build   # BUILD SUCCEEDED
xcodebuild test  -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'
# TEST SUCCEEDED — 90 passed, 0 failed
```

Trial-merged in a throwaway worktree against `origin/master` and each of #476, #487, #467, #473, #477, #482 — **all clean, no conflicting hunks**. Merged with current `master` (4 commits ahead): **TEST SUCCEEDED, 96 passed, 0 failed**.

New localized strings are present for en / fr / vi / zh-Hans, and the UI note next to the toggle now states the ownership rule alongside the existing crash/force-quit limitation.

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -2205,6 +2205,35 @@
         }
       }
     },
+    "agents.restoreOnQuit.ownership" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Only configs Quotio can verify it wrote are reverted. A config edited outside Quotio, or one pointing at a different local proxy, is left untouched. Apply an agent's setup once from here so Quotio can record it."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Seules les configurations dont Quotio peut vérifier qu'il les a écrites sont restaurées. Une configuration modifiée en dehors de Quotio, ou pointant vers un autre proxy local, reste intacte. Appliquez une fois la configuration d'un agent depuis cet écran pour que Quotio puisse l'enregistrer."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Chỉ những cấu hình mà Quotio xác minh được là do chính nó ghi mới được hoàn nguyên. Cấu hình bị chỉnh sửa bên ngoài Quotio, hoặc trỏ tới một proxy cục bộ khác, sẽ được giữ nguyên. Hãy áp dụng thiết lập của agent một lần tại đây để Quotio ghi nhận."
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "仅还原 Quotio 能够确认由自己写入的配置。在 Quotio 之外被修改的配置，或指向其他本地代理的配置，将保持不变。请在此处应用一次代理工具的设置，以便 Quotio 记录归属。"
+          }
+        }
+      }
+    },
     "agents.saveConfig" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -2118,6 +2118,93 @@
         }
       }
     },
+    "agents.restoreOnQuit" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Restore agent configs when Quotio quits"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Restaurer les configurations des agents à la fermeture de Quotio"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Khôi phục cấu hình agent khi thoát Quotio"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "退出 Quotio 时还原代理工具配置"
+          }
+        }
+      }
+    },
+    "agents.restoreOnQuit.desc" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "When Quotio quits, it removes its proxy settings from every agent it configured, so those agents go back to their own subscriptions. A backup is written before each change."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "À la fermeture, Quotio retire ses paramètres de proxy de chaque agent qu'il a configuré, afin que ces agents reviennent à leurs propres abonnements. Une sauvegarde est écrite avant chaque modification."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Khi thoát, Quotio sẽ xóa thiết lập proxy khỏi mọi agent mà nó đã cấu hình, để các agent đó quay lại dùng gói đăng ký riêng. Một bản sao lưu được ghi trước mỗi thay đổi."
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "退出时，Quotio 会从其配置过的每个代理工具中移除代理设置，让这些工具回到各自的订阅。每次修改前都会写入备份。"
+          }
+        }
+      }
+    },
+    "agents.restoreOnQuit.limitation" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Off by default. A crash or a force quit cannot be intercepted, so configs stay in place after one."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Désactivé par défaut. Un plantage ou une fermeture forcée ne peut pas être intercepté : les configurations restent alors en place."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Mặc định tắt. Không thể can thiệp khi ứng dụng gặp sự cố hoặc bị buộc thoát, nên cấu hình sẽ được giữ nguyên trong các trường hợp đó."
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "默认关闭。崩溃或强制退出无法被拦截，此时配置将保持不变。"
+          }
+        }
+      }
+    },
     "agents.saveConfig" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/QuotioApp.swift
+++ b/Quotio/QuotioApp.swift
@@ -491,6 +491,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // Stop background polling
         AtomFeedUpdateService.shared.stopPolling()
 
+        // Opt-in: hand agent configs back to their own subscriptions before the
+        // proxy they point at disappears.
+        restoreAgentConfigurationsIfEnabled()
+
         CLIProxyManager.terminateProxyOnShutdown()
         
         // Use semaphore to ensure tunnel cleanup completes before app terminates
@@ -508,6 +512,27 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             // Fallback: force kill orphan processes if stopTunnel timed out
             TunnelManager.cleanupOrphans()
             NSLog("[AppDelegate] Tunnel cleanup timed out, forced orphan cleanup")
+        }
+    }
+
+    /// Reverts agents Quotio configured back to their default (non-proxy) setup.
+    /// No-op unless the user opted in. Runs on a detached task while the main
+    /// thread waits, so it cannot deadlock against the main actor, and is capped
+    /// by a timeout so a slow filesystem can never hang termination.
+    private func restoreAgentConfigurationsIfEnabled() {
+        guard AgentQuitRestoreSettings().isEnabled else { return }
+
+        let semaphore = DispatchSemaphore(value: 0)
+        let restoreTimeout: DispatchTime = .now() + .seconds(3)
+
+        Task.detached(priority: .userInitiated) {
+            let outcome = await AgentQuitRestoreService.shared.restoreConfiguredAgents()
+            NSLog("[AppDelegate] Agent config restore on quit: \(outcome.summary)")
+            semaphore.signal()
+        }
+
+        if semaphore.wait(timeout: restoreTimeout) == .timedOut {
+            NSLog("[AppDelegate] Agent config restore on quit timed out")
         }
     }
 
--- a/Quotio/QuotioApp.swift
+++ b/Quotio/QuotioApp.swift
@@ -487,13 +487,53 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             && window.level == .normal
     }
 
+    /// Opt-in: hand agent configs back to their own subscriptions *before* AppKit
+    /// starts tearing the app down.
+    ///
+    /// The revert touches several files per agent, so it cannot be squeezed into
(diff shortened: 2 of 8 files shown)


nguyenphutrong (changes requested):
Thank you for the thoughtful implementation and focused test coverage. I verified a clean synthetic merge with current `master`; the Debug build and all 79 tests pass. I do not think the quit-time behavior is safe or complete enough to merge yet:

1. **Restoration is not guaranteed to finish before termination.** `applicationWillTerminate` starts an unstructured detached task and returns after a fixed three-second wait (`QuotioApp.swift:522-536`). After a timeout, AppKit may terminate the process while the serial loop is still restoring later agents, leaving a partially restored set of configs. Please defer the termination decision until restoration completes (for example via `applicationShouldTerminate` / `.terminateLater` and `reply(toApplicationShouldTerminate:)`), with an explicit timeout/recovery outcome rather than allowing the task to outlive the callback.

2. **`localhost` does not establish Quotio ownership.** `isConfiguredForQuotioProxy` delegates to the existing broad `isProxyConfigured` signal (`AgentQuitRestoreService.swift:47-50`), which treats any `localhost` or `127.0.0.1` endpoint as Quotio-managed. With this option enabled, quitting Quotio can therefore remove settings for a different user-managed local proxy; Factory Droid's default path can remove every localhost custom model. Please track/verify the exact configuration Quotio wrote (including detecting external changes) before modifying user files.

3. **Several supported configuration paths remain active after the restore.** Claude's `shellOnly`/`both` setup writes proxy exports to the shell profile, but quit detection/restoration only reads `~/.claude/settings.json`. Codex setup writes `auth.json` with Quotio's API key, while the default path only edits `config.toml`; Amp similarly leaves the URL-keyed entry in `secrets.json`. These cases can still leave an agent in API/proxy mode after Quotio exits. The restore needs to cover every file/storage mode written by setup while preserving pre-existing credentials.

4. **The issue's original-state/restart requirement is still missing.** Issue #128 asks to restore the original file, or remove it if Quotio created it, and to retain/reload Quotio's own agent settings on restart. This change strips known keys from the current file, never records whether the file originally existed, and does not persist/reapply the managed agent configuration on the next launch. Please either implement those semantics or explicitly narrow the issue/PR contract before claiming it closes #128.

Please also add integration-level tests for timeout/partial completion, non-Quotio localhost configurations, shell-based Claude setup, auxiliary credential files, and created-vs-pre-existing config files; the current stub tests cannot exercise these failure modes.

giaBaoJS:
Thanks — all four points were actionable. Pushed `981d452`; PR description rewritten to match.

**1. Termination is now deferred, not raced.**

`applicationShouldTerminate(_:)` returns `.terminateLater`; `AgentQuitRestoreTerminationCoordinator` (in `AgentQuitRestoreService.swift`) owns the reply. Sequence:

1. Toggle off → `.terminateNow`. Second quit while a restore is in flight → `.terminateNow`.
2. Otherwise two `@MainActor` tasks start — the serial revert, and a 5 s watchdog — and `.terminateLater` is returned.
3. Whichever finishes first calls `resolve(_:)`, guarded by a one-shot flag on the main actor, then `NSApp.reply(toApplicationShouldTerminate: true)`. Completion logs the outcome; timeout logs the agents that *did* complete, read from a lock-protected progress snapshot rather than through the loop (so a wedged revert cannot block the watchdog).
4. AppKit proceeds → `applicationWillTerminate` → proxy/tunnel teardown, unchanged.

`.terminateLater` makes AppKit wait indefinitely, so a missed reply would hang the app and force a Force Quit — worse than the original bug. The watchdog is what makes the deferral safe; both tasks capture the coordinator strongly so a dropped reference cannot swallow the reply. 5 s keeps it inside the shorter OS budget for a logout/restart quit. Restoration now runs before `CLIProxyManager.terminateProxyOnShutdown()` instead of racing it.

Re #476: `AppResetService.relaunch()` calls `exit(0)`, bypassing every termination handler including this one. That is the outcome we want — reset relaunches Quotio immediately, so agent configs should stay put. No line overlap either (#476 does not touch `QuotioApp.swift`).

**2. Ownership no longer keys off `localhost`.**

New `Quotio/Services/AgentConfigOwnership.swift`. When Quotio writes a config it records a receipt — the endpoint plus a SHA-256 fingerprint of the key it issued (raw key never persisted). At quit time the on-disk endpoint *and* key must still match. Receipt present but not matching ⇒ changed outside Quotio ⇒ left alone. No receipt ⇒ fall back to a Quotio-authored marker, and only where one is unambiguous. Neither ⇒ skip.

I enumerated what setup actually writes per agent:

| Agent | Quotio-authored in-file marker | Key evidence used by receipt |
|---|---|---|
| Claude Code | **none** — `ANTHROPIC_*` are Anthropic's own variable names | `env.ANTHROPIC_AUTH_TOKEN` |
| Codex CLI | **yes** — `model_provider = "cliproxyapi"` + `[model_providers.cliproxyapi]` (`AgentConfigurationService.swift:340-355`) | `auth.json` → `OPENAI_API_KEY` |
| Amp CLI | **none** — `amp.url` is Amp's own key | `secrets.json` → `apiKey@<amp.url>` |
| OpenCode | **yes** — `provider.quotio` / `"name": "Quotio"` (`AgentConfigurationService.swift:1198-1224`) | `provider.quotio.options.apiKey` |
| Factory Droid | **none** — `custom_models[]` entries carry no Quotio field | `custom_models[].api_key` |

So three of five agents have no reliable in-file marker. Without a receipt they are skipped rather than guessed at — including the Factory Droid case you named: its default revert filters *every* localhost `custom_models` entry, so unproven ownership now means the revert is not run at all. Cost is that a config applied by an older build has no receipt; the UI note (localized en/fr/vi/zh-Hans) says to apply setup once so Quotio can record it.

**3 and 4 — narrowed explicitly rather than over-claimed.** "Closes #128" is gone from the description; it now says it addresses the revert-on-quit half. Not implemented: original-file restore / delete-if-created (the receipt records `primaryConfigExisted`, but the revert still strips keys), and persist-and-reapply on next launch.

On the auxiliary paths — these are gaps in the shared revert path, which this PR deliberately does not modify, not in the quit orchestration:

- **Claude `shellOnly`/`both`**: `ShellProfileManager.removeFromProfile` exists (`ShellProfileManager.swift:62`) but has **no callers anywhere in the repo** — the per-agent "Default" button does not undo shell exports either. With `shellOnly` no JSON is written, so this PR reads no config and skips the agent: not made worse, not helped.
- **Codex `auth.json`**: #467 adds exactly this to `generateCodexDefaultConfig`, the path this PR calls, so it is inherited when #467 lands.
- **Amp `secrets.json`**: `generateAmpDefaultConfig` removes `amp.url` but leaves the `apiKey@<url>` entry — inert once the URL is gone, but untidy.

Happy to fold any of these in here or split them out; they touch `AgentConfigurationService.swift`, which is busy right now, so I left it untouched.

**Tests.** `AgentConfigOwnershipTests.swift` (14) runs entirely against a throwaway temp home — no real agent config is read or written. Covers: receipt proving ownership for all five agents across the `/v1` and trailing-slash forms; a foreign `http://localhost:4000` proxy **not** owned for Claude/Amp/Factory Droid; Factory Droid per-entry ownership both ways; rotated Claude token and externally replaced Codex `auth.json` key ⇒ `externallyModified`; marker fallback recognised for Codex/OpenCode but not for foreign providers; receipt round-trip with the raw key absent from the stored blob. `AgentQuitRestoreTests.swift` (11) adds: unowned agents never reach the reverter; all agents complete before the reply fires; a wedged agent hits the timeout, still replies, and reports partial progress; the reply fires exactly once when the revert lands just after the deadline; failure isolation unchanged.

```
xcodebuild ... build   # BUILD SUCCEEDED
xcodebuild test ...    # TEST SUCCEEDED — 90 passed, 0 failed
```

Trial-merged in a throwaway worktree against `origin/master` and each of #476, #487, #467, #473, #477, #482 — all clean. Merged with current `master` (4 commits ahead): TEST SUCCEEDED, 96 passed, 0 failed.

nguyenphutrong (changes requested):
The narrowed scope and receipt-based ownership are useful improvements, but three blockers remain. First, the timeout and second-quit paths permit termination without cancelling/settling the restore task, so it can keep modifying files and be killed mid-write. Second, Factory Droid ownership is proven when any entry matches, but the reused default reverter then removes every localhost model, including user-owned siblings; removal must be constrained to receipt-matched entries. Third, Claude shellOnly/both setup is not reverted, so shell exports can continue pointing at a dead Quotio proxy after quit. Please make timeout/second-quit settlement explicit, remove only proven Factory entries, and cover shell-based Claude setup. The synthetic merge builds and the full suite passes.

nguyenphutrong:
Thank you for the contribution. After review, the quit-time restore flow and its ownership and termination design are no longer aligned with Quotio's current plan. Addressing the requested changes would require a different approach, so this PR is being closed. The linked issue will remain open for future consideration.

Issue #257, opened 2026-01-31
[Bug] Refresh not working from menu bar


aldinokemal:
### What happened?

<img width="366" height="409" alt="Image" src="https://github.com/user-attachments/assets/685cbf8b-b812-4d16-a989-ebb46af5c4f2" />

data not being updated when refreh from menubar

### Steps to reproduce

Refresh from statusbar

### Expected behavior

it should make data up to date

### Severity

🟠 High - Core feature broken, no workaround

### Affected component

Menu Bar

### Provider (if applicable)

None

### CLI Agent (if applicable)

None

### macOS Version

Mac OS Latest

### Quotio Version

Latest as today (just updated)

### App Mode

Quota-Only Mode

### Logs (optional)

```shell

```

PR #486, opened 2026-08-12
fix(quota): refresh imported Cursor and Trae quotas on global refresh

### LINKED ISSUE #163: Cursor Menubar Incorrect Information
### What happened?

Image

I'm having an issue with the cursor, my quota isn't updated. I've already refreshed, but it still doesn't update. To make it up to date, I have to open the UI and refresh from there.

### Steps to reproduce

Cursor

### Expected behavior

It should updated

### Severity

### Affected component

Dashboard

### Provider (if applicable)

None

### CLI Agent (if applicable)

None

### macOS Version

MacOS 26

### Quotio Version

0.7.1

### App Mode

Full Mode (Proxy running)

### Screenshots

_No response_

### Logs (optional)

```shell

```

### Additional context

_No response_

giaBaoJS:
## Problem

Cursor quota shown in the menu bar never updates. Refreshing from the menu bar does nothing; the user has to open the main window and refresh from the **Providers** screen to get fresh numbers (#163).

## Root cause — the asymmetry

Every global refresh action funnels into one of three methods in `QuotaViewModel`, and **none of them touch Cursor or Trae**:

| Path | Cursor/Trae refreshed? |
|---|---|
| `refreshAllQuotas()` — Full Mode with proxy running | no (`// Note: Cursor and Trae removed from auto-refresh`) |
| `refreshQuotasUnified()` — Quota-Only Mode | no (`// Note: Cursor and Trae removed - require explicit scan`) |
| `refreshQuotasDirectly()` — Monitor Mode | no (its provider set omits them) |

The menu bar refresh button (`MenuActionsView` → `manualRefresh()`) and the legacy `MenuActionHandler.refresh()` (→ `refreshQuotasUnified()`) only ever reach those three, so the menu bar can never update a Cursor account.

The main window's **Providers** toolbar refresh does one extra thing:

```swift
if !modeManager.isMonitorMode {
    await viewModel.refreshAutoDetectedProviders()
}
```

`refreshAutoDetectedProviders()` iterates `AIProvider.allCases.filter { !$0.supportsManualAuth }`, which includes `.cursor` and `.trae`, and calls the scoped `refreshQuota(for:)` — the only global entry point that hits `refreshCursorQuotasInternal()`. That is exactly the "I have to open the UI and refresh from there" workaround the reporter describes.

## Fix

New `refreshImportedIDEQuotas()`, called from all three global refresh methods, refreshes the quota of Cursor/Trae accounts that are **already imported**:

- For each IDE-derived provider that currently has entries in `providerQuotas`, read the local IDE database and apply the fresh values.
- `mergeImportedIDEQuotas(fetched:into:)` merges **only into account keys that are already present**. Keys that exist only in the freshly-read data are dropped; keys the fetcher could not read this time keep their previous value rather than vanishing from the menu bar.
- If a provider has nothing imported, the fetcher is not even called, so an unscanned Cursor install is never pulled in.
- Skips a provider whose scoped refresh is already in flight, then persists via the existing `savePersistedIDEQuotas()`.

In Monitor Mode the call is placed before `discoverAccounts(merging:)` / `coordinator.finish(quotas:)`, so the refreshed values land in the Monitor snapshot too. All three methods already end with `notifyQuotaDataChanged()`, which is what drives the menu bar redraw, so the menu bar picks the new numbers up in the same pass.

## Composition with #472

#472 removes `.cursor`/`.trae` from `refreshAutoDetectedProviders()` because that path **re-imports** them from the local IDE databases wholesale, which both bypasses the explicit "Scan for IDEs" consent model (#29) and resurrects accounts the user deleted (#213). This PR is the complement and does not undo that:

- It never calls `refreshCursorQuotasInternal()` / `refreshTraeQuotasInternal()` and never goes through `refreshAutoDetectedProviders()`. It uses the same "read, then keep only the already-known account keys" shape as the scoped per-account path (`refreshQuota(for account:)`), which #472 explicitly preserves.
- Import remains gated on an explicit scan: with `providerQuotas[provider]` empty, `refreshImportedIDEQuotas()` returns without fetching.
- Deleted accounts stay deleted: after #472's `deleteAutoDetectedAccount` / `deleteMonitorAccount` clear the entry, the key is no longer in `providerQuotas`, so the merge drops it out of the freshly-read data. `testMergeImportedIDEQuotasDoesNotResurrectDeletedAccount` asserts this directly, so #472's guarantee is protected by a test.

The two PRs are independent (no shared hunks) and together give: no re-import on a generic refresh, but a working quota refresh for whatever is imported.

## Relation to #257

#257 ("Refresh not working from menu bar", Quota-Only Mode) reports the same symptom from the same code path — the menu bar refresh reaching `refreshQuotasUnified()`, which skips Cursor/Trae. It is very likely the same root cause and should be fixed by this change, but since #257 does not name a provider I am only claiming #163 here.

## Scope

Touches only `Quotio/ViewModels/QuotaViewModel.swift` — no overlap with the menu bar rewrite in #449 (`StatusBarMenuBuilder.swift`, `QuotaScreen.swift`, `Localizable.xcstrings`).

## Tests

New `QuotioTests/IDEQuotaRefreshTests.swift`:

- `testMergeImportedIDEQuotasUpdatesImportedAccount` — the stale Cursor snapshot the menu bar was showing is actually replaced.
- `testMergeImportedIDEQuotasDoesNotResurrectDeletedAccount` — an account key present in the IDE database but not imported is **not** added back (#213 / #472 guarantee).
- `testMergeImportedIDEQuotasImportsNothingWhenProviderNotScanned` — nothing is imported without an explicit scan (#29).
- `testMergeImportedIDEQuotasKeepsAccountMissingFromFetch` — a transiently unreadable IDE database does not wipe the account from the menu bar.
- `testIDEQuotaRefreshCoversBrowserAuthProviders` — the refresh set is exactly the IDE-derived (`usesBrowserAuth`, non-manual-auth) providers.

```
xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build   # ** BUILD SUCCEEDED **
xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug \
  -destination 'platform=macOS'                                                  # ** TEST SUCCEEDED **
```

Fixes #163

### DIFF
--- a/Quotio/ViewModels/QuotaViewModel.swift
+++ b/Quotio/ViewModels/QuotaViewModel.swift
@@ -198,7 +198,9 @@ final class QuotaViewModel {
     // MARK: - IDE Quota Persistence Keys
 
     private static let ideQuotasKey = "persisted.ideQuotas"
-    private static let ideProvidersToSave: Set<AIProvider> = [.cursor, .trae]
+    /// Providers whose quota is imported from a local IDE database and persisted
+    /// separately (Cursor, Trae). Also drives `refreshImportedIDEQuotas()`.
+    nonisolated static let ideProvidersToSave: Set<AIProvider> = [.cursor, .trae]
 
     /// Key for tracking when auth files last changed (for model cache invalidation)
     static let authFilesChangedKey = "quotio.authFiles.lastChanged"
@@ -750,6 +752,11 @@ final class QuotaViewModel {
         providerQuotas[.openRouter] = await openRouter
         providerQuotas[.amp] = await amp
         providerQuotas = providerQuotas.filter { !$0.value.isEmpty }
+
+        // Refresh (not re-import) Cursor/Trae accounts the user already scanned in,
+        // before the snapshot is persisted, so the menu bar refresh shows the same
+        // numbers as the Providers screen (#163).
+        await refreshImportedIDEQuotas()
 
         monitorAccounts = await coordinator.discoverAccounts(merging: providerQuotas)
         removeDisabledMonitorQuotas()
@@ -1594,6 +1601,10 @@ final class QuotaViewModel {
             async let clinePass: () = refreshClinePassQuotasInternal()
 
             _ = await (antigravity, openai, copilot, claudeCode, glm, warp, kiro, clinePass)
+
+            // Refresh (not re-import) Cursor/Trae accounts the user already scanned in,
+            // so the menu bar refresh shows the same numbers as the Providers screen (#163).
+            await refreshImportedIDEQuotas()
         }
 
         checkQuotaNotifications()
@@ -1636,6 +1647,10 @@ final class QuotaViewModel {
         async let clinePass: () = refreshClinePassQuotasInternal()
 
         _ = await (antigravity, codex, copilot, claudeCode, glm, warp, kiro, clinePass)
+
+        // Refresh (not re-import) Cursor/Trae accounts the user already scanned in,
+        // so the menu bar refresh shows the same numbers as the Providers screen (#163).
+        await refreshImportedIDEQuotas()
 
         checkQuotaNotifications()
         pruneMenuBarItems()
@@ -2012,6 +2027,70 @@ final class QuotaViewModel {
         
         for provider in autoDetectedProviders {
             await refreshQuotaForProvider(provider)
+        }
+    }
+    
+    /// Merge freshly read IDE quota data into the accounts that are already imported.
+    ///
+    /// Cursor and Trae accounts only exist once the user has explicitly run
+    /// "Scan for IDEs" (issue #29), and they can be deleted (issue #213). A global
+    /// refresh must therefore update the quota of the accounts that are currently
+    /// imported without importing any account key that is not already present —
+    /// otherwise a deleted account would resurrect on the next refresh.
+    ///
+    /// - Parameters:
+    ///   - fetched: everything the IDE fetcher could read from the local database.
+    ///   - existing: the account keys currently imported into `providerQuotas`.
+    /// - Returns: `existing`, with fresh data applied to the keys it already holds.
+    ///   Keys present only in `fetched` are dropped; keys the fetcher could not read
+    ///   this time keep their previous value instead of disappearing from the UI.
+    nonisolated static func mergeImportedIDEQuotas(
+        fetched: [String: ProviderQuotaData],
+        into existing: [String: ProviderQuotaData]
+    ) -> [String: ProviderQuotaData] {
+        guard !existing.isEmpty else { return existing }
+
+        var merged = existing
+        for (accountKey, quota) in fetched where existing[accountKey] != nil {
+            merged[accountKey] = quota
+        }
+        return merged
+    }
+
+    /// Refresh the quota of already-imported IDE-derived accounts (Cursor, Trae).
+    ///
+    /// This is the piece the global refresh actions were missing: `refreshAllQuotas`,
+    /// `refreshQuotasUnified` and `refreshQuotasDirectly` all skip Cursor/Trae, so the
+    /// menu bar refresh never updated them (issues #163, #257). Unlike the per-provider
+    /// re-import in `refreshQuota(for:)`, this only touches account keys that are
+    /// already imported, so it never re-imports from the IDE databases.
+    private func refreshImportedIDEQuotas() async {
+        var didRefresh = false
+
+        for provider in Self.ideProvidersToSave {
+            let existing = providerQuotas[provider] ?? [:]
+            // Nothing imported for this provider: respect the explicit-scan consent
+            // model (issue #29) and do not pull anything in.
+            guard !existing.isEmpty else { continue }
+            // A scoped refresh for this provider is already in flight; let it win.
+            guard !activeRefreshProviders.contains(provider) else { continue }
+
+            let fetched: [String: ProviderQuotaData]
+            switch provider {
+            case .cursor:
+                fetched = await cursorFetcher.fetchAsProviderQuota()
+            case .trae:
+                fetched = await traeFetcher.fetchAsProviderQuota()
+            default:
+                continue
+            }
+
+            providerQuotas[provider] = Self.mergeImportedIDEQuotas(fetched: fetched, into: existing)
+            didRefresh = true
+        }
+
+        if didRefresh {
+            savePersistedIDEQuotas()
         }
     }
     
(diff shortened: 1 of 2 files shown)


nguyenphutrong (changes requested):
Thanks for fixing the missing Cursor/Trae global-refresh path and for preserving the explicit-import boundary. I verified the PR as a clean synthetic merge into the current `master`: the Debug build succeeds and all 76 tests pass.

One concurrency blocker remains in `refreshImportedIDEQuotas()`. The method snapshots `existing` and checks `activeRefreshProviders` only before awaiting the IDE fetcher. `QuotaViewModel` is `@MainActor`, but the actor is reentrant across that await, so another scoped refresh can begin while the global IDE refresh is in flight. The global refresh then unconditionally writes a merge based on its pre-await snapshot, which can overwrite the newer scoped result even though the comment says the scoped refresh should win.

This also weakens the PR's composition guarantee with #472: once Cursor/Trae deletion is available, a deletion performed while the fetch is awaiting can remove and persist the account, after which this write restores it from the captured `existing` dictionary. In other words, `testMergeImportedIDEQuotasDoesNotResurrectDeletedAccount` covers a static merge, but not the reentrant state change that can actually resurrect it.

Please make the IDE refresh participate in the refresh-coordination slot for the full fetch/write operation, and merge against/revalidate the current imported keys after the await so state removed during the fetch cannot be restored. Please also add a regression test for state changing while an IDE fetch is suspended (at minimum, deletion; ideally also a concurrent scoped refresh). The remaining observations are non-blocking: `git diff --check` reports trailing whitespace after the new helper, and the three global paths still have duplicated call-site plumbing.

giaBaoJS:
Thanks — the reentrancy analysis was right, and the previous test could not have caught it. Pushed `6f73e14`.

**1. The slot is now held for the whole fetch/write.** `refreshImportedIDEQuotas()` is split per provider, and each provider claims `activeRefreshProviders` through the existing `beginBatchRefresh(providers:)` / `endBatchRefresh(providers:)` pair before the fetch and releases it with `defer` after the write. `.cursor`/`.trae` are absent from the provider sets of all three global paths (`refreshQuotasDirectly` `QuotaViewModel.swift:586-589`, `refreshAllQuotas` `:1582`, `refreshQuotasUnified` `:1628-1630`), so the acquisition is always disjoint from the enclosing batch and cannot deadlock or be refused by it.

Because `beginScopedRefresh` (`:247-250`) inserts into that same set, a scoped refresh can no longer *start* midway through the IDE fetch — so the "who wins" question is resolved by exclusion rather than by a late check. The two orderings are now: scoped refresh already running → `beginBatchRefresh` fails and the global IDE refresh yields to it; global IDE refresh running → the scoped one is rejected at its guard before doing any I/O. That is the same coordination policy every other provider already follows, and it is honest in the UI: `isRefreshBlocked(for:)` reads the same set, and it drives `.disabled(...)` on the per-account refresh button in both `QuotaScreen.swift:717` and `StatusBarMenuBuilder.swift:992`, while `refreshingProviders` drives the spinner (`QuotaScreen.swift:140`). Previously those controls stayed enabled during the IDE fetch.

**2. State is revalidated after the await.** The write no longer uses the pre-await snapshot: `providerQuotas[provider]` is re-read *after* the suspension point, and `mergeImportedIDEQuotas` is applied to that value, so only keys still present at write time are touched. If the provider entry is gone entirely (last imported account deleted) the write is skipped. `savePersistedIDEQuotas()` now runs in the same synchronous step as the write instead of once after the loop, so the persisted snapshot cannot capture a state from a concurrent mutation between providers. Provider iteration is sorted, since `Set` order is unspecified and the sequence of suspension points needs to be deterministic.

**3. Regression tests that actually interleave.** Added `IDEQuotaRefreshReentrancyTests`, backed by a `#if DEBUG` injectable fetch (`ideQuotaFetchHookForTesting`) so a test can suspend *inside* the fetch on a continuation, mutate the view model from the test body exactly as a user action would, then resume:

- `testGlobalRefreshDoesNotResurrectAccountDeletedWhileFetchIsSuspended` — two imported Cursor accounts, one deleted while the fetch is suspended, fetch returns both. Asserts the deleted key is absent from `providerQuotas` **and** from the decoded `persisted.ideQuotas` payload, and that the surviving account still gets the fresh quota.
- `testGlobalRefreshDoesNotRestoreProviderRemovedWhileFetchIsSuspended` — the whole `.cursor` entry removed mid-fetch must stay removed.
- `testScopedRefreshIsBlockedWhileImportedIDEFetchIsSuspended` — asserts `isRefreshBlocked(for:)`/`isRefreshing(provider:)` are true *while suspended*, that `refreshQuota(for: .cursor)` issued at that moment is a no-op, and that a second overlapping `refreshImportedIDEQuotas()` performs no fetch (`fetchCount == 1`).
- `testGlobalRefreshDoesNotFetchWhenNothingIsImported` — with nothing imported, no slot, no fetch, no import (issue #29 consent).

The tests save and restore `persisted.ideQuotas` around each case so the host app's real data is untouched. I confirmed they are not vacuous: reverting only the production method to the previous snapshot-based semantics makes the first three fail (`** TEST FAILED **`, the fourth still passes, as expected since it never reaches the await).

**End-user behaviour re-verified:** all three global paths still end in `notifyQuotaDataChanged()` (`:770`, `:1612`, `:1657`); `QuotioApp.swift:70-79` observes it and calls `updateStatusBar()` + `rebuildMenuInPlace()`; the title reads `viewModel.providerQuotas` (`QuotioApp.swift:93`, `:117`) and the dropdown reads it at `StatusBarMenuBuilder.swift:121` and `:227`. So the merged Cursor/Trae value lands in `providerQuotas` before the notification that redraws the menu bar, which is the #163/#257 symptom. `beginBatchRefresh`/`endBatchRefresh` also post that notification, so the in-flight and released states are reflected too.

**Non-blocking points:** `git diff --check` is clean now. On the duplicated call-site plumbing — I trimmed the repeated comments to one line, but left the three call sites as-is: they are three separate entry points with different provider sets, different modes and different orderings relative to `coordinator.finish(quotas:)`, and collapsing them is a refactor wider than this fix. Happy to do it in a follow-up if you'd like it in this PR.

**Verification:** `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` succeeds with no new warnings; `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS'` — `** TEST SUCCEEDED **`, 0 failures, run three times; `MonitorRuntimeTests.testCoordinatorCoalescesConcurrentRefreshForProvider` passed in every run.

nguyenphutrong (approved):
The actor-reentrancy blocker is resolved. Imported-IDE refresh holds the provider slot across fetch/write and re-reads current keys after suspension, preventing scoped-refresh overwrite and account resurrection. Deterministic interleaving tests cover deletion, overlap, persistence, and no-import consent; the synthetic merge and full suite pass.

PR #485, opened 2026-08-12
refactor(quota): carry explicit bucket kinds from quota fetchers

giaBaoJS:
## What changed

Reworked per review: the classification is no longer derived from `ModelQuota.name`. Each fetcher now declares the bucket kind at construction, from the payload field the bucket came from.

```swift
nonisolated enum QuotaWindow: String, Codable, Sendable, CaseIterable { case session, daily, weekly, monthly }
nonisolated enum QuotaBilling: String, Codable, Sendable { case subscription, paidOverage }

// on ModelQuota
var window: QuotaWindow?    // nil = the provider states no window
var billing: QuotaBilling?  // nil = the provider gives no billing signal
```

Both are optional, and `nil` means unknown rather than a guess. `nil` billing stays in the subscription group, so an unclassified bucket is never dropped from `.sessionOnly` totals. Because `ProviderQuotaData` is cached to disk, optional fields also keep old caches decodable (covered by a test).

The total-usage math moved to a pure `QuotaUsageCalculator`, and `totalUsagePercent` now takes `[ModelQuota]` instead of `(name, percentage)` tuples so callers stop discarding the kinds.

## Scope correction (please read before the diff)

Investigating point 2 showed the PR's premise was wrong, so I retitled it from `fix` to `refactor` and removed the Codex claim.

- Codex never emits a paid-overage `ModelQuota`. Paid credits are the analytics row built in `CodexUsageMapper.analytics` (`CodexUsageMapper.swift:124-143`), not a bucket consumed by `totalUsagePercent`. `codex-extra` in the old literal set is dead — no fetcher has ever produced it. The suffix rule is gone.
- With today's producers this changes **no displayed number**. The old literal set already covered the only two paid buckets that carry a percentage (`extra-usage`, `on-demand`). `grok-extra-usage`, `devin-extra-balance` and `factory-extra-balance` were on the wrong side of the split, but all three emit `percentage: -1` and `aggregateModelPercentages` filters those, so totals were unaffected.

What the change buys is that billing/window semantics are grounded at the producer instead of in a display slug, and that #474 has a correct window signal to select on.

## Bucket-kind inventory (every producer, as of this branch)

| Producer | Bucket | window | billing | Evidence |
|---|---|---|---|---|
| ClaudeCodeQuotaFetcher | `five-hour-session` | session | subscription | `info.fiveHour` |
| | `seven-day-weekly` / `-sonnet` / `-opus` | weekly | subscription | `info.sevenDay*` |
[...]

PR #439, opened 2026-07-20
Expand Monitor mode quota providers

nguyenphutrong:
## Summary

- add Monitor quota support for Devin, Grok, OpenRouter, and Factory Droid
- upgrade the existing GLM integration for Z.ai-compatible endpoints and richer quota buckets
- add progress, amount, and status metric presentations without breaking saved snapshots
- integrate provider discovery, secure API-key accounts, periodic refresh, menu filtering, and quota rendering
- keep Factory Droid focused on Standard/Core limits and Extra Balance without a redundant Extra Usage status row

## User impact

Monitor mode can track more provider accounts independently of Local Proxy. API-key providers use the existing credential vault, while supported local apps and CLIs can be discovered from their native credentials.

## Validation

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -destination 'platform=macOS'` — 41 tests passed
- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build`
- `git diff --check`

Live-account refresh and light/dark-mode visual checks remain pending.

PR #487, opened 2026-08-12
feat(refresh): add opt-in adaptive quota refresh cadence

### LINKED ISSUE #172: Refresh quota by checking agent client activated status
### Problem Statement

希望刷新频率可以支持根据用户的对模型的使用情况切换

发现目前只能设置固定的时间进行数据更新，但是在使用的时候往往用量数据变化的很快，往往会出现 10 分钟后用量从 0 - 100 的情况；
同理若是发现用户当前没有使用，可以退避到固定的时间刷新或者指数退避时间刷新

### Proposed Solution

可以参考 

Image

reference：https://github.com/f-is-h/Usage4Claude

### Alternative Solutions

_No response_

### Feature Category

Quota Tracking Enhancement

🟡 Medium - Nice to have

### Use Case

_No response_

### Mockups / Examples

_No response_

### Additional Context

_No response_

### Checklist

giaBaoJS:
## Motivation

Today the refresh interval is a single fixed value, and that value has to serve two opposite situations. As the reporter puts it in #172:

> 发现目前只能设置固定的时间进行数据更新，但是在使用的时候往往用量数据变化的很快，往往会出现 10 分钟后用量从 0 - 100 的情况；同理若是发现用户当前没有使用，可以退避到固定的时间刷新或者指数退避时间刷新

Translated: while you are actively coding, usage can jump from 0% to 100% inside one 10 minute window, so the default cadence shows stale numbers exactly when they matter. Conversely, when you are not using anything, a short cadence keeps polling provider APIs all night for identical data.

This PR adds an opt-in **Adaptive refresh** mode alongside the existing fixed cadences. It is **off by default**, so nothing changes for existing users unless they enable it.

## Behaviour

- **Fast interval** = the cadence already chosen in Settings. Adaptive mode can only slow refreshes down, never speed them up beyond what the user picked.
- **Active** (usage moved recently): keep refreshing at the fast interval.
- **Idle**: after a grace window of two fast intervals, the interval doubles per poll, capped at a **30 minute ceiling** (never below the chosen cadence, so the 15 min cadence yields 15 → 30 min).
- **Activity resumes**: the interval snaps straight back to the fast cadence on the poll that observes the change.

Backoff schedule with a 1 minute cadence: `60 → 60 → 120 → 240 → 480 → 960 → 1800 → 1800 …` seconds.

## Signal choice

The activity signal is **the quota usage numbers Quotio already fetches**, compared between two consecutive snapshots. Per row the value is the typed `QuotaMetricPresentation` where the provider supplies one (`.progress(used:)`, `.amount(value:)` — the Double-valued dollar/credit metrics used by Amp, OpenRouter, Factory Droid, Devin and GLM), otherwise the legacy `used` counter, otherwise the consumed percentage.

Why this one, rather than proxy request traffic (`RequestTracker`) or a running agent CLI process:

- It works in **Full and Monitor** modes. `RequestTracker` is only fed by `ProxyBridge` while the local proxy is running, so it is blind in Monitor mode, which is exactly where most quota-only users live.
- **Remote proxy mode is explicitly out of scope.** `refreshAllQuotas()` skips the local fetchers there and `refreshData()` does not populate `providerQuotas`, so there is no local activity signal to read. In Remote mode every automatic poll reports `.skipped`, the cadence stays exactly fixed, and the Settings toggle is hidden rather than offered as a setting that would do nothing.
- It requires **no new data source and no process spying** — it is computed from `providerQuotas`, which the refresh loop already produces.
- It measures the thing the issue is actually about: quota consumption moving. Traffic through the local proxy is a proxy for that (pun intended) and misses usage made directly by an agent that is not routed through Quotio.
- Refresh timestamps are deliberately excluded from the comparison, so re-fetching unchanged data never registers as activity.

The trade-off is that activity is detected at poll resolution: after a long idle period the first poll that observes new usage is what resets the cadence. Since the ceiling is 30 minutes and the reset is immediate on that poll, this keeps the worst case well inside what a fixed 10–15 minute cadence already gives, while staying fast for the whole active session.

## Implementation

- `Quotio/Services/AdaptiveRefreshPlanner.swift` (new)
  - `AdaptiveRefreshBounds` — tunables derived from the chosen cadence (fast interval, grace, multiplier, ceiling).
  - `AdaptiveRefreshPlanner.nextInterval(current:timeSinceLastActivity:bounds:)` — **pure, timer-free** cadence decision, the unit under test.
  - `QuotaUsageSignature` — usage snapshot used as the activity signal.
  - `AdaptiveRefreshTracker` — small `Sendable` value type that folds snapshots into the next interval. Still no timers, so it is directly testable.
- `RefreshSettingsManager` gains `adaptiveRefreshEnabled` (persisted, default `false`) and `adaptiveBounds`. Toggling it reuses the existing `onRefreshCadenceChanged` callback so the refresh loop restarts.
- `QuotaRefreshOutcome` (`.sample` / `.skipped` / `.failed`) is returned by `refreshAllQuotas()`, `refreshQuotasUnified()` and `refreshQuotasDirectly()`. Only `.sample` is evidence about the user: `.skipped` (coalesced early return, no local quota source, refresh not due) leaves both the cadence and the comparison baseline untouched, and `.failed` (empty result, or a provider the Monitor coordinator flagged) recovers the cadence to the user's interval so a backed-off delay never survives a failure.
- `QuotaViewModel`: the three auto-refresh loops (`startAutoRefresh`, `startQuotaOnlyAutoRefresh`, `startQuotaAutoRefreshWithoutProxy`) now sleep via `sleepUntilNextRefresh(fixedNanoseconds:)` and feed the **outcome of the refresh they awaited** into the tracker. With the toggle off, that helper sleeps on `RefreshCadence.intervalNanoseconds` exactly as before.
- `refreshData()` keeps a handle to the quota refresh it schedules (it still does not await it, so the management-API results reach the UI immediately). `performAutomaticRefresh()` awaits that handle before recording, so the automatic loop observes the refresh whose result it records instead of the previous poll's snapshot.
- `monitorStatus(for:)` uses the interval actually in effect for its staleness threshold, so a deliberately backed-off account is not mislabelled "outdated".
- **Coalescing and manual refresh untouched**: the adaptive interval is clamped to `>= fastInterval`, so it can never poll harder than fixed mode, and per-provider coalescing in `MonitorRefreshCoordinator` (#439/#440) still de-duplicates in-flight work. `manualRefresh()` is not modified.
- Settings UI: a toggle plus help text in the existing `RefreshCadenceSettingsSection`, hidden when the cadence is `manual` (nothing to adapt). New strings localized in en, fr, vi, zh-Hans.

## Test evidence

`QuotioTests/AdaptiveRefreshPlannerTests.swift` — pure cadence and signature logic:

- activity → fast interval (at zero idle and inside the grace window)
- increasing idle → `[60, 120, 240, 480, 960, 1800, 1800, …]` with the ceiling enforced
- ceiling never below the chosen cadence; a cadence longer than the ceiling is never shortened
- adaptive never returns an interval below the chosen cadence, swept over 0–2 h of idle time
- tracker driven by real `ProviderQuotaData` snapshots: backs off while usage is unchanged, snaps back to the fast interval the moment usage moves, `reset()` returns to fast
- signature ignores `lastUpdated`, tracks percentage when `used` is unavailable, and detects added accounts/models
- **typed metrics**: an Amp-style `.amount` balance (`percentage: -1`, no `used`) moving 20 → 12.5 and an OpenRouter-style `.progress` row moving 4.25 → 4.75 USD at an unchanged integer percentage both register as activity
- **outcome gating**: `.skipped` changes neither interval nor baseline; `.failed` pulls a 30 minute backoff back to the retry cadence and never overwrites the baseline
- fixed `RefreshCadence` interval constants are unchanged

`QuotioTests/AdaptiveRefreshIntegrationTests.swift` — runtime behaviour through `QuotaViewModel`:

- **ordering**: suspends inside the quota refresh and proves nothing is observed while the fetch is in flight, then that the recorded signature is the value the refresh wrote
- a coalesced (`.skipped`) poll does not back off even after the clock passes the idle grace window
- a `.failed` poll recovers a ceiling-level interval to the user's cadence
- **opt-out parity**: with the toggle off no tracker is armed for any cadence and the loop's computed sleep equals `cadence.intervalNanoseconds`, including after ten refreshes that would otherwise have backed off

```
** BUILD SUCCEEDED **

** TEST SUCCEEDED **   (89 tests)
```

## Interaction with other open PRs

Trial-merged in a throwaway worktree (since removed). Pairwise onto this branch, **#480, #486, #472, #474, #485 and #476 all merge clean**. Cumulatively, #480 + #486 + #472 + #474 + #476 on top of this branch merges clean, builds, and the full suite passes.

#485 conflicts with #474 in `Quotio/QuotioApp.swift`, but that conflict reproduces on `master` with only those two applied and involves no file this PR touches, so it is between those two PRs and independent of this one.

Closes #172

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -18821,6 +18821,64 @@
         }
       }
     },
+    "settings.refresh.adaptive" : {
+      "extractionState" : "stale",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Adaptive refresh"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Actualisation adaptative"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Làm mới thích ứng"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "自适应刷新"
+          }
+        }
+      }
+    },
+    "settings.refresh.adaptive.help" : {
+      "extractionState" : "stale",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Keeps the interval above while usage is changing, then doubles it while you are idle, up to 30 minutes. Never refreshes more often than the interval above."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Conserve l’intervalle ci-dessus tant que la consommation évolue, puis le double pendant l’inactivité, jusqu’à 30 minutes. N’actualise jamais plus souvent que l’intervalle ci-dessus."
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Giữ nguyên khoảng thời gian ở trên khi mức sử dụng thay đổi, sau đó tăng gấp đôi khi bạn không hoạt động, tối đa 30 phút. Không bao giờ làm mới thường xuyên hơn khoảng thời gian ở trên."
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "用量变化时保持上面的刷新间隔；闲置时按倍数退避，最长 30 分钟。刷新频率不会高于上面设置的间隔。"
+          }
+        }
+      }
+    },
     "settings.refresh.cadence" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/Quotio/Models/MenuBarSettings.swift
+++ b/Quotio/Models/MenuBarSettings.swift
@@ -380,7 +380,8 @@ final class RefreshSettingsManager {
     
     private let defaults = UserDefaults.standard
     private let refreshCadenceKey = "refreshCadence"
-    
+    private let adaptiveRefreshKey = "adaptiveRefreshEnabled"
+
     /// Current refresh cadence
     var refreshCadence: RefreshCadence {
         didSet {
@@ -388,13 +389,32 @@ final class RefreshSettingsManager {
             onRefreshCadenceChanged?(refreshCadence)
         }
     }
-    
+
+    /// Opt-in adaptive cadence: keep `refreshCadence` while usage is moving and
+    /// back off progressively while it is idle (issue #172). Off by default, so
+    /// the cadence above behaves exactly as before unless enabled.
+    var adaptiveRefreshEnabled: Bool {
+        didSet {
+            guard adaptiveRefreshEnabled != oldValue else { return }
+            defaults.set(adaptiveRefreshEnabled, forKey: adaptiveRefreshKey)
+            onRefreshCadenceChanged?(refreshCadence)
+        }
+    }
+
+    /// Backoff bounds for adaptive refresh, derived from the chosen cadence.
+    /// `nil` when there is no automatic refresh to adapt (manual cadence).
+    var adaptiveBounds: AdaptiveRefreshBounds? {
+        guard let seconds = refreshCadence.intervalSeconds else { return nil }
+        return AdaptiveRefreshBounds(cadenceInterval: seconds)
+    }
+
     /// Callback when refresh cadence changes (for ViewModel to restart timer)
     var onRefreshCadenceChanged: ((RefreshCadence) -> Void)?
     
     private init() {
         let saved = defaults.string(forKey: refreshCadenceKey) ?? RefreshCadence.tenMinutes.rawValue
         self.refreshCadence = RefreshCadence(rawValue: saved) ?? .tenMinutes
+        self.adaptiveRefreshEnabled = defaults.bool(forKey: adaptiveRefreshKey)
     }
 }
 
--- a/Quotio/Services/AdaptiveRefreshPlanner.swift
+++ b/Quotio/Services/AdaptiveRefreshPlanner.swift
@@ -1,0 +1,219 @@
+//
+//  AdaptiveRefreshPlanner.swift
+//  Quotio - Adaptive Quota Refresh Cadence
+//
+//  A fixed refresh cadence has to trade freshness against API pressure: a
+//  10 minute cadence can miss a quota going 0% -> 100% while the user is
+//  actively coding, while a 1 minute cadence keeps hammering provider APIs
+//  overnight when nothing is happening (issue #172).
+//
+//  Adaptive refresh keeps the user's chosen cadence as the *fast* interval and
+//  only backs off when no usage activity is observed, doubling the interval up
+//  to a documented ceiling and snapping back to the fast cadence the moment
+//  usage moves again.
+//
+//  Everything in this file is pure and timer-free so the cadence decision can
+//  be unit tested without waiting on wall-clock time.
+//
+
+import Foundation
+
+// MARK: - Bounds
+
+/// Tunables for the adaptive backoff curve.
+nonisolated struct AdaptiveRefreshBounds: Sendable, Equatable {
+    /// Interval used while the user is active. This is the cadence picked in
+    /// Settings, so adaptive mode never refreshes *more* often than the fixed
+    /// mode would - it can only slow down.
+    var fastInterval: TimeInterval
+
+    /// How long the fast interval is held after the last observed activity
+    /// before backoff starts.
+    var idleGrace: TimeInterval
+
+    /// Growth factor applied per idle refresh.
+    var multiplier: Double
+
+    /// Upper bound for the backed-off interval.
+    var maxInterval: TimeInterval
+
+    /// Ceiling for the whole backoff schedule, in seconds.
+    static let defaultMaxInterval: TimeInterval = 1800  // 30 minutes
+
+    /// Doubling backoff.
+    static let defaultMultiplier: Double = 2
+
+    init(
+        fastInterval: TimeInterval,
+        idleGrace: TimeInterval,
+        multiplier: Double = AdaptiveRefreshBounds.defaultMultiplier,
+        maxInterval: TimeInterval = AdaptiveRefreshBounds.defaultMaxInterval
+    ) {
+        self.fastInterval = fastInterval
+        self.idleGrace = idleGrace
+        self.multiplier = multiplier
+        self.maxInterval = maxInterval
+    }
+
+    /// Bounds derived from the cadence the user picked in Settings.
+    ///
+    /// The fast interval is the chosen cadence, the grace window is two of
+    /// those intervals (so a single quiet poll does not immediately slow things
+    /// down), and the ceiling is 30 minutes - never shorter than the chosen
+    /// cadence, so picking the 15 minute cadence still yields 15 -> 30 minutes.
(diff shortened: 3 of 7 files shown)


nguyenphutrong (changes requested):
Thanks for the well-isolated planner, opt-in setting, localization, and extensive pure-state tests. I verified a clean synthetic merge into the current `master`: the Debug build succeeds and all 83 tests pass.

The planner itself looks sound, but the runtime integration has three blockers:

1. In `startAutoRefresh()`, `recordAdaptiveObservation()` does not observe the refresh that just completed. `refreshData()` starts `refreshAllQuotas()` in a new unstructured `Task` and returns immediately, so the observation at the next line snapshots the old `providerQuotas` while the actual quota refresh is still running. This makes Full/Remote behavior one poll late at best and can record a false idle observation while a refresh is in flight. Please make the automatic loop await the quota refresh whose result it records, and add an integration test that suspends the fetch to prove the observation happens afterward.

2. All three loops record an observation unconditionally even when no successful quota sample was produced. The batch refresh methods can return early when another refresh owns a provider, and transient provider failures commonly retain the previous snapshot. Both cases currently look exactly like “unchanged usage,” so the tracker backs off—and can keep a previously backed-off 30-minute delay—precisely when the app needs to retry. Please distinguish a completed usable sample from skipped/failed refreshes; failures should not be evidence of inactivity and should recover to an appropriate retry cadence.

3. `QuotaUsageSignature` ignores the typed metric values now used by several providers. It chooses only `model.used` or `model.usedPercentage`; `.amount(value:)` rows from Amp, OpenRouter, Factory Droid, and Devin generally have `percentage == -1`, so every balance becomes the constant `101` and real consumption is invisible. `.progress(used:limit:)` should also use its Double-valued presentation instead of relying on the optional legacy Int. Please include the relevant `QuotaMetricPresentation` values and add amount/progress regression tests.

There is also an unsupported cross-mode claim: in Remote mode `refreshAllQuotas()` skips local quota fetchers, and `refreshData()` does not populate `providerQuotas`, so this signal is empty rather than working identically. Please either provide a real Remote activity source or scope/disable the feature there and adjust the description.

Non-blocking: `testDisabledAdaptiveModeUsesExactlyTheFixedCadence` only verifies enum constants; it does not exercise disabled runtime behavior or timer parity.

giaBaoJS:
Thanks for the precise review — all three runtime blockers are addressed in `ee2135d`.

**1. The loop now awaits the refresh it records.** `refreshData()` still returns without awaiting the quota refresh (the auth files / usage stats / API keys must reach the UI without waiting on provider round-trips, and `manualRefresh()` drives `refreshAllQuotas()` itself), but it now keeps a handle to the task it schedules. The automatic loop goes through `performAutomaticRefresh()`, which awaits that handle before `recordAdaptiveObservation(_:)` reads `providerQuotas`. The handle and `errorMessage` are re-read *after* `refreshData()` returns rather than from a pre-await snapshot.

`AdaptiveRefreshIntegrationTests.testAutomaticRefreshObservesTheQuotaRefreshItStarted` proves the ordering: it suspends inside the quota refresh, asserts nothing has been observed while the fetch is in flight, then releases it and asserts the recorded signature is the *new* value. Reverting just the await and restoring the old ordering makes it fail exactly as described:

```
AdaptiveRefreshIntegrationTests.swift:108: XCTAssertNil failed:
  "QuotaUsageSignature(entries: ["claude|user@example.com|sonnet": 10.0])"
  - the observation must not be recorded while the quota refresh is still in flight
AdaptiveRefreshIntegrationTests.swift:118: XCTAssertEqual failed:
  ("Optional(10.0)") is not equal to ("Optional(55.0)")
  - the observation must snapshot the refresh that just completed, not the previous poll
```

**2. "No sample" is now a typed outcome, not an inference.** `QuotaRefreshOutcome` is `.sample` / `.skipped` / `.failed`, returned by `refreshAllQuotas()`, `refreshQuotasUnified()` and `refreshQuotasDirectly()`. `.skipped` covers the `beginBatchRefresh` early return, a poll where no quota refresh was due, and the "proxy is running" branch of `startQuotaAutoRefreshWithoutProxy()`. `.failed` covers an empty result and any provider the Monitor coordinator flagged (`issues[provider]` is set for both the failure and the partial-result cases in `MonitorRefreshCoordinator.refresh`, which is exactly where the previous snapshot is retained). In the tracker, `.skipped` changes neither the interval nor the comparison baseline, and `.failed` resets the interval to the user's cadence so a backed-off 30 minute delay cannot survive a failure. Neither overwrites the baseline, so the next real sample is compared against the last real sample.

**3. `QuotaUsageSignature` reads the typed metric.** `QuotaUsageSignature.consumption(of:)` now takes `.progress(used:)` and `.amount(value:)` from `QuotaMetricPresentation` and only falls back to the legacy `used`/`usedPercentage` for `.status` and untyped rows. Regression tests cover both: an Amp-style `.amount` balance (built with `percentage: -1`, no `used` — previously the constant `101`) moving 20 → 12.5, and an OpenRouter-style `.progress` row moving 4.25 → 4.75 USD at an unchanged integer percentage. A tracker-level test confirms a backed-off cadence snaps back when only an `.amount` value moves. `.balance` falls while `.spent` rises; the signature only tests inequality, so both read as activity.

**Remote mode.** You're right that the claim was unsupported, and I've scoped the feature out rather than inventing a signal: `refreshAllQuotas()` and `refreshQuotasUnified()` return `.skipped` in Remote mode, so the tracker never observes and the cadence stays exactly fixed there, and the Settings toggle is hidden in Remote mode instead of being offered as a setting that does nothing. I deliberately did not build it on the remote `usageStats` because #480 is actively reworking that field in `refreshData()`. The PR description is updated to match.

**Non-blocking point.** `testDisabledAdaptiveModeUsesExactlyTheFixedCadence` is replaced. The constants check is now `testFixedCadenceConstantsAreUnchanged`, and two new runtime tests exercise disabled behaviour through the view model: no tracker is armed for any cadence and the loop's computed sleep equals `cadence.intervalNanoseconds`, including after ten refreshes that would have backed a tracker off.

**Verification.** Debug build succeeds; `xcodebuild test -destination 'platform=macOS'` passes (89 tests, up from 83).

**Trial merges** (throwaway worktree, then removed). Pairwise onto this branch: #480, #486, #472, #474, #485 and #476 all merge clean. Cumulatively, #480 + #486 + #472 + #474 + #476 on top of this branch merges clean, builds, and the full suite passes. #485 conflicts with #474 in `Quotio/QuotioApp.swift`, but that conflict reproduces on `master` with only those two applied and no file of mine involved, so it is between those two PRs and independent of this one.

nguyenphutrong (changes requested):
The awaited observation, typed metric, skipped outcome, and Remote-mode issues are addressed. One production failure path remains: Full-mode refreshAllQuotas()/refreshQuotasUnified() call completedRefreshOutcome() without provider-level results, and it returns .sample whenever the aggregate providerQuotas map is nonempty. A provider fetch can fail while unrelated or retained rows keep that map populated, so the failed poll is still recorded as unchanged usage and increases backoff. The current failure test injects .failed rather than exercising production classification. Please propagate per-provider success/failure through the Full-mode batch and add real partial/total failure regression tests. The synthetic merge builds and the full suite passes.

nguyenphutrong:
Thank you for the contribution. This branch now conflicts with master, and the adaptive refresh cadence proposed here is no longer part of the current refresh plan. Bringing it up to date would require both conflict resolution and a different product direction, so this PR is being closed. The linked issue will remain open for future prioritization.

PR #490, opened 2026-08-12
feat: add optional YubiKey secret vault

bulyaki:
## Summary
- keep the existing non-interactive Keychain behavior for unsigned/local builds
- add optional YubiKey PIV-backed storage for Quotio-owned secrets
- migrate legacy Keychain secrets only after successful encryption
- expose YubiKey detection, setup, selection, and refresh in Settings

## Validation
- `xcodebuild -project Quotio.xcodeproj -scheme Quotio -configuration Debug build` ✅
- Build completed with local signing and no Developer ID dependency.

## Scope
- Qwen changes are excluded as obsolete.
- Developer ID, packaging, version, and release-script changes are excluded.
- Physical YubiKey provisioning and UI flow require manual verification.


nguyenphutrong (changes requested):
Thank you for splitting this into a focused contribution. The envelope design is a reasonable direction, and the synthetic merge builds cleanly, but this is not safe to merge yet.

1. **The setup flow can silently destroy an existing PIV key.** `beginProvisioning` unconditionally runs `piv keys generate ... 9d` and then writes a new certificate to slot `9d` (`YubiKeySecretVault.swift:116-118`). Current `ykman` replaces the private key and certificate in an occupied slot without a confirmation/preflight; the old private key is irrecoverable. The preceding `change-management-key --protect --generate` also replaces the key's administrative credential and can break other PIV tooling. Please detect existing slot/management state and refuse destructive provisioning by default, with an explicit, accurately worded confirmation for any operation that changes it.

2. **Monitor credential updates target the wrong backend after migration.** `compareAndSwapMonitorCredential` first calls `getMonitorCredential` (`KeychainHelper.swift:155-163`). With the vault enabled, that read migrates the item and deletes the Keychain copy (`:320-330`), but the CAS then always calls `SecItemUpdate` on the deleted Keychain item (`:164-174`). Consequently, any refresh-token/access-token rotation for a loaded Quotio monitor credential returns `credentialWriteFailed` and is not persisted. The compare-and-swap update must write to the active backend.

3. **A vault read failure is incorrectly treated as “no envelope.”** `YubiKeySecretVault.read` returns `nil` for a missing file, malformed/corrupt data, a cancelled PIN prompt, a wrong key, and decryption failure. `readData` then falls back to any legacy Keychain value, overwrites the envelope, and deletes the Keychain item. This can silently roll a newer vault credential back to a stale Keychain copy. Use the existing `contains` check (or a typed read result) to distinguish absence from failure, and never overwrite an existing unreadable envelope. Before deleting the legacy item, verify a full encrypt/decrypt round trip rather than only a successful public-key encryption/write.

4. **The new Settings UI bypasses the project's localization contract and can block the UI.** All new labels/status/help text in `SettingsScreen.swift:1889-1934` are hardcoded English. Also, `refresh()` calls `provisionableDevices()` synchronously from `onAppear`, while that method launches `ykman` and waits with `Process.waitUntilExit()` (`YubiKeySecretVault.swift:80-99`); a slow or stuck process freezes Settings. Please localize these strings and perform device/process discovery off the main actor.

Verification on the current `master` tip:
- Clean synthetic merge
- Debug build succeeded
- Full test suite succeeded (71 tests)
- Embedded Debug signature contains the smart-card entitlement and passes strict code-sign verification

The existing tests run only with the vault disabled, so please add regression coverage for active-backend CAS, absent-vs-unreadable envelopes, and non-destructive migration. Physical YubiKey provisioning still needs the manual validation already noted in the PR.

bulyaki:
Thanks for the detailed review — all four blocking items are addressed, plus the coverage you asked for and the physical validation that was still outstanding. Two commits on top of the original: `0894fb0` and `843bf4d`.

**1. Destructive provisioning.** `beginProvisioning` is gone, and with it the Terminal hand-off. Setup now runs through `YubiKeySecretVault.preflight(_:)`, which parses `ykman piv info` before anything is written, and a sheet that shows what will happen to *this* key.

The important part is what "empty" is allowed to mean. PIV key metadata only exists on firmware ≥ 5.3, so on older keys a bare private key in 9D is invisible to `piv info`. Reading "not listed" as "empty" would be exactly the silent destruction you flagged, so slot state is a three-way `.empty / .occupied / .unknown`, and `.unknown` is treated as destructive. Provisioning is refused by default and the confirmation names the specific consequence — overwriting an existing key vs. replacing the management key — rather than a generic "are you sure".

The management key change is likewise conditional: `preflight` detects whether it is already PIN-protected or derived (`has_protected_key` in ykman's terms), and provisioning only supplies one when ykman will actually ask.

The PIN is collected in-app and written to `ykman` on **stdin**, never as an argument — argv is world-readable via `ps` for any same-uid process. Worth noting the original comment claiming credentials "are never passed through Quotio" no longer holds, and I removed it rather than leave it stale.

**2. CAS targeting the wrong backend.** `compareAndSwapMonitorCredential` now dispatches on `monitorCredentialBackend(vaultEnabled:)` and writes through the vault when the vault is active, instead of unconditionally calling `SecItemUpdate` on an item the migrating read had already deleted.

**3. Read failure treated as absence.** `read` is replaced by a typed `readResult` returning `.success / .absent / .unreadable`, and legacy migration is gated on `.absent` only. Before the Keychain copy is deleted, a full encrypt→decrypt round trip must succeed and compare equal to the original — a successful public-key write alone is no longer enough.

While testing on hardware I found this cuts deeper than the read path. A caller that cannot read a secret sees only `nil`, and cannot distinguish "never stored" from "key unplugged". `CLIProxyManager.init` does exactly that: read, and on `nil` mint a fresh UUID and save it. With the vault enabled, simply launching Quotio without the key inserted permanently overwrote the envelope holding the real management key — no leftover state required. `843bf4d` guards the single choke point every such path funnels through: `saveData` refuses to overwrite an envelope it could not read. `.absent` stays writable so first storage and migration still work, and costs only a file-existence check, so no write pays for a decrypt it does not need.

That commit also wires up `shouldMigrateLegacy`, which had no production caller — its test only proved it agreed with itself.

**4. Localization and blocking the UI.** 45 keys added across `en`, `fr`, `vi`, `zh-Hans`; every one verified to resolve in all four `.lproj` bundles via `xcrun xcstringstool compile` rather than by eye, since `NSLocalizedString` returns the key on a miss and looks plausible in the UI. Count and index specifiers are `%lld` — `String(format:)` bridges Swift `Int` as 64-bit, so `%d` reads the wrong half.

Every `ykman` invocation now runs inside `Task.detached(priority: .userInitiated)`; nothing calls `provisionableDevices()` or `availableIdentities()` on the main actor, and discovery moved from `onAppear` to `.task`.

**Coverage.** Added for active-backend CAS, absent-vs-unreadable envelopes, non-destructive migration, preflight parsing (fixtures rendered through ykman's own `pretty_print`, so the shape can't drift from the tool), and the overwrite guard. Full suite: 76 passing.

**Physical validation** — done, on a YubiKey 5C NFC (firmware 5.4.3), and it earned its keep. Provisioning succeeded, the slot was correct, every test passed, and Settings still said "not configured": `availableIdentities()` compared `kSecAttrTokenID` against the driver bundle ID, but macOS reports the token *instance* — `com.apple.pivtoken:48B9336C…`. The comparison could never match, so every PIV identity was invisible and the feature was inert on real hardware while looking healthy in CI. Fixed in `isPIVToken`, with the observed instance ID as a test fixture.

The same session surfaced a UI incoherence worth mentioning: after setup, Settings could show "Not configured" and "YubiKey set up — choose the key below" three lines apart, with no picker in between. The section now switches on a single `VaultStatus`, so contradictory rows are unrepresentable rather than merely avoided.

---

**One design question I'd like your call on before merge.** The slot is provisioned `--pin-policy always`, so *every* decrypt raises a PIN prompt. In practice that means a prompt per credential read, including background monitor refreshes. It bit the test suite first: `testMonitorCredentialVaultAddsRotatesAndDeletesAPIKeys` writes and reads through whatever backend is active, so on a machine with a configured key it drove real hardware — six PIN prompts, 90 seconds, and a failure. I've skipped that test when the vault is enabled (with a message saying how to run it), since forcing the Keychain path would mean switching off a real security setting from a test.

The skip fixes the suite; it doesn't fix the ergonomics for users. `once` would prompt a single time per card session and is the usual choice for a decrypt-often key, at the cost of a weaker guarantee if the machine is left unlocked with the key inserted. `always` is the stronger posture and defensible, but it should be a deliberate choice rather than a default nobody revisits — and if you want it kept, the docs should say plainly that each stored secret costs a PIN entry. Happy to change it either way; it only affects newly provisioned keys.

nguyenphutrong (approved):
The prior security blockers are resolved: provisioning uses conservative slot preflight and explicit destructive acknowledgement, process work is off the main actor, monitor CAS writes the active backend, unreadable envelopes cannot trigger migration/overwrite, and legacy secrets are deleted only after an encrypt/decrypt equality check. The synthetic merge and targeted MonitorRuntimeTests pass. I recommend retaining pin-policy always as the secure default; document the repeated-prompt behavior and consider once only as a future explicit lower-assurance option.

PR #493, opened 2026-08-13
feat(menu-bar): stack Claude 5h and weekly quota

afu-it:
## Summary

- show Claude's 5-hour session quota above its weekly quota in one compact menu bar item
- add a persistent **Stack Claude 5h and weekly quota** toggle under Settings > Menu Bar, enabled by default
- preserve the existing aggregate display when the toggle is off, and leave other providers unchanged
- keep the status item narrow with a 22-point height, localized accessibility labels, and balanced click-edge spacing

## Why

Claude exposes separate 5-hour and 7-day limits, but the current menu bar item reduces them to one aggregate percentage. Seeing both windows at a glance makes the menu bar substantially more useful while avoiding an additional status item.

Top row: 5-hour session. Bottom row: weekly quota.

Compact Claude 5-hour and weekly quota in the macOS menu bar

## Testing

- `xcodebuild test -project Quotio.xcodeproj -scheme Quotio -configuration Debug -destination 'platform=macOS'`
- 75 tests passed, including model selection, missing-window fallback, and compact fitting-size coverage
- manually verified the menu bar layout and live toggle on macOS

### DIFF
--- a/Quotio/Localizable.xcstrings
+++ b/Quotio/Localizable.xcstrings
@@ -16828,6 +16828,35 @@
         }
       }
     },
+    "settings.menubar.stackClaudeQuotaWindows" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Stack Claude 5h and weekly quota"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Empiler les quotas Claude sur 5 h et hebdomadaire"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Xếp chồng quota Claude 5 giờ và hàng tuần"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "堆叠显示 Claude 5 小时和每周配额"
+          }
+        }
+      }
+    },
     "settings.networkAccess" : {
       "extractionState" : "stale",
       "localizations" : {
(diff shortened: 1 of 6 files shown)


nguyenphutrong (changes requested):
Thanks for the focused implementation and coverage. I reviewed the current head as a clean synthetic merge into `master` (`8d9a0f69`); the Debug test action passes with all 293 tests. Two issues need to be addressed before merge:

1. **A missing window is announced as a percentage rather than unavailable data.** `quotaWindowsAccessibilityLabel` feeds `-1` through `formatPercentage`, producing `--%` (`StatusBarManager.swift` around lines 295–307). VoiceOver will read punctuation/percent instead of the localized no-data state. This conflicts with the unknown-quota accessibility contract already used by `RingProgressView` and `StatusBarMenuBuilder`, which expose `quota.noDataYet`. Please format each accessibility value semantically—localized no-data text for a missing window, percentage only for a known value—and add coverage for the one-window case.

2. **The spacing change affects every status item, not only the stacked Claude layout.** The PR removes `StatusBarQuotaView`'s existing 4-point horizontal padding and reduces the outer container padding from 4 to 1 (`StatusBarManager.swift` around lines 82–90 and 216–223). This also changes non-Claude providers, the normal aggregate layout when the toggle is off, and the default icon/click edge, despite the PR saying those paths remain unchanged. Please keep the existing global spacing and make any compaction local to the stacked content, or explicitly revise the scope and provide manual/automated coverage for all affected status-item layouts and click targets.

Non-blocking integration note: #474 also changes menu-bar quota selection and conflicts with this PR in `QuotioApp.swift`. Once #474 is updated, the combined behavior needs an explicit precedence between an exact per-item bucket selection and the global Claude stacking option rather than resolving the conflict textually.

PR #497, opened 2026-08-14
fix(menubar): render native template image to eliminate glow and unfocused blur

tqhit:
### Summary

- Fixes the menu bar unfocused blur, washed-out text, and vibrancy glow when the menu bar window is inactive on secondary/inactive displays.
- Renders the status bar quota content directly into a native template `NSImage` via `ImageRenderer` on `NSStatusBarButton.image`, eliminating the need for `NSHostingView` subview layers.
- Uses `targetBackingScaleFactor` derived from `NSScreen.screens` to dynamically preserve maximum resolution across multi-monitor setups without scale drops when switching focus.
- Removes intermediate `StatusBarContainerView` class and restores canonical AppKit event handling.
- Keeps typography font sizes, weights, and provider icon assets completely unchanged from master for minimal diff.

### DIFF
--- a/Quotio/QuotioApp.swift
+++ b/Quotio/QuotioApp.swift
@@ -95,6 +95,7 @@ final class AppBootstrap {
         statusBarManager.updateStatusBar(
             items: quotaItems,
             colorMode: menuBarSettings.colorMode,
+            quotaDisplayMode: menuBarSettings.quotaDisplayMode,
             isRunning: hasQuotaData,
             showMenuBarIcon: menuBarSettings.showMenuBarIcon,
             showQuota: menuBarSettings.showQuotaInMenuBar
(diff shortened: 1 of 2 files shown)


nguyenphutrong:
@codex review

nguyenphutrong:
Thanks for the fix! Since this change affects the menu bar’s visual rendering, could you please add before/after screenshots if possible? It would be especially helpful to compare the active and inactive states on a secondary display, and both colored and monochrome modes if available.

nguyenphutrong:
Thanks! I tested the old version and this PR side by side. While both look nearly identical when active, the old rendering becomes visibly blurrier when the display loses focus, whereas this PR keeps the text and provider icons noticeably sharper. This confirms that the fix is useful and addresses a real issue on my setup.

Before merging, could you please address the existing review feedback, particularly preserving VoiceOver accessibility, re-rendering when the Used/Remaining display mode changes, retaining adaptive appearance in colored mode, and maintaining an adequate click target/padding? A short before/after recording of the focus transition would also be helpful for documenting the fix.

Thank you for working on this!

nguyenphutrong:
Merged—thank you! I also opened #503 as a follow-up to refresh the status item when its appearance or highlight state changes.

PR #526, opened 2026-09-04
fix: keep Claude Code credentials read-only

owenmead:
Supersedes #465, reduced to the scope you asked for there, and re-ported onto the layered architecture.

## What changed since #465

The background adopter is gone entirely. That removes the incompatibility you identified — with #466 making external reads non-interactive, an unauthorized read returns `nil` rather than prompting, so the one-time copy could never happen for the users it targeted — and it removes all three behavioural defects with it, since disabled-state handling, the multi-account conflation and the `includeMonitorCredentials` bypass all lived in the adopter.

What remains is the part you said was worth preserving: staying read-only toward Claude Code.

## Re-ported, not rebased

`ClaudeCodeQuotaFetcher.swift` no longer exists. The Clean Architecture migration replaced it with `QuotioInfrastructure/Quota/ClaudeQuotaFetcher.swift` — 675 lines down to 358, with credential loading extracted into `LocalClaudeQuotaCredentialLoader` and `CompositeClaudeQuotaCredentialLoader`. Every hunk here was rewritten against that structure rather than rebased through it.

The restructure also surfaced a **third** write path the original PR did not cover: `ClaudeDesktopCredentialReader` now produces a `ClaudeQuotaCredential` that went through the same refresh branch as everything else.

## The boundary

Reading is not the boundary — refreshing is, because a refresh spends a token the CLI also holds. `claude` guards refresh with a cross-process lock, detects sibling races, and marks a refresh token it did not spend itself as dead on `invalid_grant`, so an unsynchronized refresh from Quotio can sign the user out of Claude Code.

Three paths were doing that, and all three now read only:

| Path | Was | Now |
|---|---|---|
| `CompositeClaudeQuotaCredentialLoader.persist` | refreshed the `Claude Code-credentials` keychain item and compare-and-swapped it back | branch deleted |
| `LocalClaudeQuotaCredentialLoader.persist` | refreshed and rewrote `~/.claude/.credentials.json` in place | refuses CLI-owned paths |
| `ClaudeDesktopCredentialReader.credential()` | fed the shared refresh branch | marked read-only |

An expired access token yields no fresh quota until the CLI renews it itself, which is the correct outcome for an observer. Files under `~/.cli-proxy-api/` and accounts in Quotio's own vault still refresh as before.

A read-only credential that fails authentication falls back to cached data rather than reporting `isForbidden`, which would ask the user to sign in to Quotio for a credential Quotio does not manage.

## Ownership is carried, not implied

You noted the previous tests didn't cover the behavioural boundaries. Rather than hardcode `allowRefresh: false` at call sites, ownership now rides on the credential itself.

`ClaudeQuotaCredential.allowsRefresh` is the single gate the fetcher and both loaders consult — one place to reason about, and impossible for a new call site to forget. `ClaudeCredentialOwnership` decides it for auth files by the directory they live in, honouring `CLAUDE_CONFIG_DIR` the way the CLI does.

Symlinked auth files are never refreshed: a `claude-*.json` entry under `~/.cli-proxy-api` pointing at `~/.claude/.credentials.json` would otherwise classify as ours and spend the CLI's token through the link. AGENTS.md requires auth files never be followed through a symlink destination, so this fails closed.

## Coverage

`ClaudeCredentialOwnershipTests` (13 tests) covers the CLI file, proxy-owned files, the `CLAUDE_CONFIG_DIR` override (including that it *relocates* ownership rather than adding to it), blank-value fallback, the `~/.claude-backup` prefix trap, path traversal, and both symlink cases.

Beyond classification, it pins the behaviour itself: an expired CLI token issues no token request, a 401 on a CLI credential does not trigger a refresh and does not surface `isForbidden`, the local loader leaves `~/.claude/.credentials.json` byte-identical after a `persist` call, the composite never compare-and-swaps the keychain item — and, as a control, that a proxy-owned credential *still* refreshes and *still* reports `isForbidden`.

## On the original framing

#465 was titled around stopping repeated keychain prompts. **#466 already fixed that** — with reads non-interactive, a wiped ACL produces a silent `nil` instead of a dialog. Claiming that fix here would be wrong, so the title and rationale describe only what this actually does: not spending the CLI's single-use refresh token.

## Verification

Re-ported onto `a001010`.

- `swift build --package-path Packages/QuotioCore` — clean
- `xcodebuild -scheme Quotio -configuration Debug -destination 'platform=macOS' build` — **BUILD SUCCEEDED**
- `swift test --package-path Packages/QuotioCore` — 551 tests
- `./scripts/check_architecture.sh` — passed

`AgentDetectionAdapterTests.testForceRefreshInvalidatesSixtySecondCache` fails on my machine but passes in CI, and it fails identically on unmodified `master` with this branch stashed — it reads local agent installations, so it is environment-dependent and unrelated to this change.


owenmead:
Both Codex findings addressed in `80acd15`.

**P1 — symlinked auth files (valid, and it defeated the whole point of the PR).** Confirmed empirically: `standardizingPath` leaves a symlink unresolved while `resolvingSymlinksInPath` resolves it, so a `claude-*.json` entry under `~/.cli-proxy-api` pointing at `~/.claude/.credentials.json` classified as ours and would have spent the CLI's refresh token through the link. AGENTS.md:139 also requires that auth files never be followed through a symlink destination, so this now fails closed: any symlinked auth file is treated as CLI-owned regardless of where it points, and the resolved path is used for directory classification. Two tests cover it — a link into the CLI directory, and a link that stays inside the proxy directory.

**P2 — `isForbidden` on read-only auth failures (valid).** The file path returned `isForbidden: true`, which asks the user to sign in to Quotio for a credential Quotio does not manage, while the keychain path returned cached data. Both now fall back to cache.

Full suite passes with the branch for #527.

zheli:
I need this. The prompt popup is really annoying

owenmead:
Re-ported onto the layered architecture (`a001010`) — `ClaudeCodeQuotaFetcher` no longer exists, so this was rewritten against `ClaudeQuotaFetcher` and the two credential loaders rather than rebased through the migration. Mergeable, `verify` green.

The restructure surfaced a third write path the original PR missed: `ClaudeDesktopCredentialReader`'s credential fed the same refresh branch as everything else. All three are read-only now, gated on a single `ClaudeQuotaCredential.allowsRefresh` rather than per-call-site flags.

Body updated to match. No rush on my end — flagging it's ready whenever you get to it.

owenmead:

PR #527, opened 2026-09-04
feat: improve Card and Ring quota layouts

owenmead:
Supersedes #449, re-ported onto the layered architecture with all four review issues addressed.

## Re-ported, not rebased

Every file this touched was moved and renamed by the Clean Architecture migration:

| #449 touched | Now |
|---|---|
| `Quotio/Services/StatusBarMenuBuilder.swift` | `QuotioPresentation/MenuBar/StatusBarMenuRenderer.swift` |
| `Quotio/Views/Screens/QuotaScreen.swift` | `QuotioPresentation/Screens/QuotaScreen.swift` |
| `Quotio/Services/Antigravity/AntigravityQuotaFetcher.swift` | split — `ModelQuota` → `QuotaMetric` in Domain |

The menu also moved to snapshot rendering in between. The layout work was rewritten against that structure rather than rebased through it, and placed to respect the new boundaries: `isUnlimitedUsage` sits with the other `QuotaMetric` presentation helpers in `QuotaPresentation.swift`, and `QuotaMetricWindow` / `RingSlotArrangement` with the shared Presentation components — Domain stays free of display rules.

## 1. Window captions no longer come from the display name

You were right that `windowCaption` could not be correct: `Session` is produced by Claude's `five-hour-session`, by Codex's variable window of up to 24 hours, and by `zai-session` alike, so a single "4h window" caption was wrong for most of them.

`QuotaMetricWindow` reads the duration from the raw metric identity, and only when that identity states one — `five-hour-*` → 5h, `seven-day-*`/`*weekly` → 7d, `*daily` → 24h. Everything else omits the caption rather than guessing, which covers `codex-session`, `zai-session` and the Antigravity groups. Calendar-variable `monthly` is deliberately omitted too.

This required plumbing the raw identity through `RingSlotData` and `ModelBadgeData`, which previously carried only display names.

## 2. Unknown quotas are no longer rendered as measurements

Since #449 you added `ringPercent` and `RingProgressView.unknownPercent`, and guards in `menuPercentText` / `menuStatusColor`. **This PR builds on those rather than introducing a parallel mechanism** — the earlier version's own sentinel handling is dropped in favour of yours.

What remained: `RingSlotBody` and the Card row still needed to render the placeholder themselves (both draw the ring with `showLabel: false` so the number can be sized for the layout), and `menuHeroText` needed the same precedence. "off" is gone with it — as you said, nothing here signals an enabled state, so asserting one was unfounded; the string is now `quota.state.unavailable`.

Standalone metrics are the exception and are handled as one: Amp/OpenRouter balances and Grok status set the sentinel *deliberately* and carry their value in `formattedUsage`, so that value is read before the sentinel is consulted. Otherwise a balance that is right there on screen would render as "unavailable".

## 3. Unlimited metrics with usage are no longer labelled "unused"

`QuotaMetric.isUnlimitedUsage` carries the distinction, and every hero helper checks usage before the `>= 100` branch, so "123 used" stays "123 used". "unused" is reserved for a genuinely zero-used window.

One thing the review didn't mention but which was necessary: the Card call site passes `formattedUsage: model.presentation == nil ? nil : model.formattedUsage`, which suppresses usage for exactly the metrics being mislabelled — Cursor on-demand reports `used`/`limit` with no typed presentation. The row then falls through to its `used`/`limit` branch, which renders only when `limit > 0`, so a metric with `used: 123, limit: nil` displays no usage at all where `formattedUsage` would have said "123 used". Fixing the precedence alone would not have fixed the reported bug. That line is still on `master` today.

## 4. Rings occupy stable semantic columns

`RingSlotArrangement` pins Session and Weekly to fixed positions and leaves a placeholder when one is absent, so an account without a Session window no longer slides Weekly into the first column. Semantic columns are only reserved for providers that use that vocabulary — a provider reporting Chat/Completions/Premium still lays out positionally instead of growing two empty columns.

The dashboard and the menu share this arrangement rather than each implementing its own, so the two views cannot disagree about the same account.

Rows wrap at four columns. To be clear about what this is and is not: `LazyVGrid` already wraps a fifth metric on its own, so master loses nothing today. A fixed `Grid` does not wrap, so `RingSlotArrangement.rows` restores that behaviour explicitly. It keeps this redesign from introducing a regression for providers like Kiro, which append both a bonus and a base metric per breakdown entry — it is not fixing an existing defect.

Dashboard columns are a fixed width, so the separate `Grid` instances line up with each other across account cards. Each ring carries an accessibility label naming the metric, its value and its window, and the menu ring shows the metric name again — without it a row of rings gives no way to tell Session from Weekly from Extra.

## Coverage

28 tests in `QuotaMetricWindowTests` and `RingSlotPresentationTests`, covering each of the four issues: the ambiguous-session case, the unlimited-with-usage case, the Weekly-without-Session layout case, the standalone-sentinel case, the more-than-four-metrics wrap, and the accessibility descriptions.

## Verification

Re-ported onto `a001010`.

- `swift build --package-path Packages/QuotioCore` — clean
- `xcodebuild -scheme Quotio -configuration Debug -destination 'platform=macOS' build` — **BUILD SUCCEEDED**
- `swift test --package-path Packages/QuotioCore` — 566 tests
- `./scripts/check_architecture.sh` — passed

`AgentDetectionAdapterTests.testForceRefreshInvalidatesSixtySecondCache` fails on my machine but passes in CI, and it fails identically on unmodified `master` with this branch stashed — it reads local agent installations, so it is environment-dependent and unrelated to this change.

I have not re-shot the light/dark screenshots against the new layout; happy to if useful.


owenmead:
All five Codex findings addressed in `1c97e21`.

**P1 — standalone values hidden by the sentinel (valid; a regression I introduced).** Amp/OpenRouter balances, Grok status, Devin and Factory set `percentage: -1` deliberately and carry the real value in `presentation`, so the new sentinel check replaced live values with "unavailable". Standalone metrics are now read before the sentinel is consulted, in the ring, the card and `menuHeroText`. Fixing `heroText` alone wasn't sufficient — `captionText` and the accessibility label went through the same branch, which the first test run caught.

**The three menu-ring findings share one cause: I fixed the dashboard ring and left the menu ring untouched** — despite that being exactly what the original review of #449 called out. Rather than patch it separately and let the two drift again, the arrangement is now extracted into `RingSlotArrangement` and shared, so the menu and the dashboard cannot disagree about the same account.

- *Semantic roles:* the menu now uses the same arrangement, so Weekly stays in its own column when Session is absent.
- *`-1` centre label:* `menuDisplayPercent` preserves the sentinel and the ring is drawn with `showLabel: false`, so the label rendered a literal "-1". It now renders the placeholder itself.
- *Metric names:* restored in the menu ring, plus a per-ring accessibility label naming the metric, its value and its window.

**P2 — truncation beyond four metrics (valid, and pre-existing in #449 rather than introduced by the rebase).** `min(max(count, 3), 4)` capped the row and dropped the remainder, where the old `LazyVGrid` wrapped. Rows now wrap at four columns; a test asserts every metric survives the layout.

Full suite passes with the branch for #526.

owenmead:
Re-ported onto the layered architecture (`a001010`). Both files this touched were moved and renamed by the migration, and the menu moved to snapshot rendering in between, so this was rewritten against the new structure. Mergeable, `verify` green.

Two things worth calling out since you reviewed the predecessor:

- Issue 2 is largely **yours** now. `RingProgressView.unknownPercent`, `menuPercentText` and the `menuStatusColor` guard all landed on master since #449, so I dropped this PR's parallel versions and built on yours.
- I corrected an overstated claim in the body: I had implied master drops metrics past the fourth ring. It doesn't — `LazyVGrid` wraps them. The wrapping logic here exists because a fixed `Grid` doesn't, so it prevents a regression rather than fixing one.

Happy to re-shoot light/dark screenshots if that would help the layout review.

owenmead:

PR #531, opened 2026-09-11
fix(ci): actually run the architecture check

owenmead:
The architecture check has never run in CI, and has been reporting success the whole time.

## What's happening

`report_matches` invokes `rg` inside an `if` condition:

```bash
if matches="$(rg --line-number --glob '*.swift' "${pattern}" "$@")"; then
```

Because the command sits in a condition, `set -e` cannot fire, and *any* non-zero exit is read as "no matches found". `rg` exits 1 for no match — but also 2 on error, and 127 when it isn't installed.

**The `macos-15` runner image does not ship ripgrep.** So every `verify` job has produced thirteen `command not found` lines on stderr, matched nothing, and finished with:

```
Architecture check passed.
```

Exit 0. A clean result for a run that inspected zero files.

You can see it directly in this PR's own history: the first push failed with

```
Architecture check requires ripgrep (rg), which was not found on PATH.
##[error]Process completed with exit code 2.
```

That is the guard working — it is the first time the runner has been honest about not having the tool.

## Why it matters

The boundaries this script defines have been unenforced since it was added. Domain importing SwiftUI, Presentation reaching for `FileManager`, a reintroduced `static let shared` — none of it would have failed a build. The rules were documented in AGENTS.md and checked by a script that silently agreed with everything.

I found this while preparing #526/#527: the script reported "passed" locally without ripgrep installed, which was suspicious enough to look into. I had initially assumed CI was fine and said so in those PRs; it wasn't.

## The fix

**Script, failing closed on both paths.** A preflight `command -v rg` guard exits 2 with an actionable message, matching the convention already used for a missing source directory. And `report_matches` now distinguishes rg's exit codes — only 1 means "ran, found nothing"; 0 is a violation, and anything higher exits 2 naming the check that could not run.

**Workflow.** The guard alone would convert a silent pass into a hard CI failure, so ripgrep is installed before the check runs.

## Does enabling it break anything?

No. Current `master` passes all thirteen checks once they actually execute — verified locally with ripgrep present. No source changes are needed alongside this.

## Verification

| Scenario | Before | After |
|---|---|---|
| Normal run, rg present | `passed`, exit 0 | `passed`, exit 0 |
| rg absent (i.e. CI, until now) | 13× `command not found`, **`passed`, exit 0** | clear message, **exit 2** |
| rg present but exits 2 | silently treated as clean | names the failing check, exit 2 |
| Planted `static let shared` in Presentation | violation, exit 1 | violation, exit 1 |

The stub case used a shim `rg` that exits 2, so the preflight guard passes and the new exit-code branch is what runs. The last row confirms detection is unchanged:

```
Architecture violation: First-party singleton declaration remains in production source
  .../QuotioPresentation/Components/ArchProbe.swift:2:enum ArchProbe { static let shared = 1 }
Architecture check failed with 1 violation group(s).
```

`HOMEBREW_NO_AUTO_UPDATE=1` keeps the install step from triggering a full `brew update`. Happy to pin a version or swap to a prebuilt action if you'd prefer.


owenmead:

